PR #12, opened 2022-10-24
fix i18n translator with a proper parser

### LINKED ISSUE #4: i18n translator may not work as expeceted
```golang

// GetLang get language from header
func GetLang(ctx *gin.Context) i18n.Language {
	acceptLanguage := ctx.GetHeader(constant.AcceptLanguageFlag)
	switch i18n.Language(acceptLanguage) {
	case i18n.LanguageChinese:
		return i18n.LanguageChinese
	case i18n.LanguageEnglish:
		return i18n.LanguageEnglish
	default:
		return i18n.DefaultLang
	}
}

```

**Accept-Language**  in some popular browser may like `zh,en;q=0.9,en-US;q=0.8,zh-CN;q=0.7,zh-TW;q=0.6,la;q=0.5,ja;q=0.4,id;q=0.3,fr;q=0.2`,

in above code logic , it will always fallback to `DefaultLang`

Fity:
Add a Accept-Language parser to fix #4 

parser code from [here](https://siongui.github.io/2015/02/22/go-parse-accept-language/)

### DIFF
--- a/internal/base/handler/lang.go
+++ b/internal/base/handler/lang.go
@@ -4,17 +4,71 @@ import (
 	"github.com/gin-gonic/gin"
 	"github.com/segmentfault/answer/internal/base/constant"
 	"github.com/segmentfault/pacman/i18n"
+	"sort"
+	"strconv"
+	"strings"
 )
+
+type LangQ struct {
+	Lang string
+	Q    float64
+}
+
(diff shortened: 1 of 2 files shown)


Fity:
Thanks for your reply, let us close this pr.

PR #13, opened 2022-10-24
update(pkg): make functions in dir more clear

k8scat:


joyqi (commented):
I like these changes. The naming is more consistent and the code is more readable. We'll consider to merge this pull request, if these codes meet our requirements.

PR #17, opened 2022-10-25
Fix RouteNode type error

ppchart:
image

### Problem Description:
`interface RouteNode extends RouteObject`
The original code `RouteNode` directly inherits `RouteObject`, and an error will be reported: An interface can only extend the object type or the intersection of object types using static known members. ts(2312)

### Problem causes:
`RouteObject` is a dynamic type: `IndexRouteObject | NonIndexRouteObject`
It is not a definite static type and therefore cannot be directly inherited

### Solution:
Declare certain types for `IndexRouteObject` and `NonIndexRouteObject` respectively: `IndexRouteNode` and `NonIndexRouteNode`
Re-declare the dynamic type: `type RouteNode = IndexRouteNode | NonIndexRouteNode`

---------------------------------

### 问题描述：
`interface RouteNode extends RouteObject`
原代码 `RouteNode` 直接继承了 `RouteObject`，会报错：接口只能扩展使用静态已知成员的对象类型或对象类型的交集。ts(2312)

### 问题原因：
`RouteObject` 是一个动态类型：`IndexRouteObject | NonIndexRouteObject`
它不是一个确定的静态类型，因此不能够被直接继承

### 解决方案：
对 `IndexRouteObject` 和 `NonIndexRouteObject` 分别声明确定的类型：`IndexRouteNode` 和 `NonIndexRouteNode`
再声明动态类型：`type RouteNode = IndexRouteNode | NonIndexRouteNode`


ppchart:
@robinv8 can we approve it or give some suggestions for it

robinv8:
@ppchart Strange, I don't see a type error, I wonder if it's caused by an external cause?

ppchart:
I found the reason:
- This issue does not appear when I use pnpm to install dependencies, and the installed react-router-dom/react-router version is: v6.4.0, because we have pnpm lock file
- This issue occurs when I use npm to install dependencies, and the react-router-dom/react-router version is: v6.4.3, because we not have package lock file

**When react-router@6.4.1 -> react-router@6.4.2 it updates RouteObject to dynamic type**

I found this relevant PR as follows:
[fix: Strengthen route typings](https://github.com/remix-run/react-router/commit/540f94b0f53335c1f0043bbf97606c79277cf820
)

It's means we have three solutions:
1. So maybe we can upload the package lock file to avoid this issue I'm having
2. Or update react-router version and merge this PR(There may be a better solution after the upgrade, not necessarily the current PR)
3. we can also choose to disallow the use of npm, in which case this PR can also be closed

ppchart:
@robinv8 ........ You can't merge this PR without updating the version of react-router-dom
OK, I commited new PR to update react-router-dom
https://github.com/answerdev/answer/pull/59

Issue #41, opened 2022-10-28
Change password for default admin does not work


freemindcore:
Thanks for the good work!
I think I hit a bug. The UI says new password saved, but it only accept the old password. Please help check.

PR #42, opened 2022-10-28
Fix bug #41 Change password does not work

geekymv:


robinv8 (approved):
Thanks for the contribution

PR #153, opened 2023-01-04
support zhTW

den13501:


haitaoo:
For the purpose of minimising the size of the package file, it is better to keep only English and Simplified Chinese as the built-in languages, and introduce other languages by dynamic loading.

> There was a bug with the dynamic loading of language files in the previous version, which has now been fixed.

Thanks!

PR #197, opened 2023-02-10
fix: 在删除标签时，判断相反，语义推测应该是同义词

lcgash:
rt, 发现一个 BUG


haitaoo:
@lcgash 

The function is set up so that **"when there is a post using a tag, you can't just delete that tag"** and you need to un-reference the tag first. So, the current description text is fine.

However, this feature does have bugs at the moment, including the handling of synonyms when deleting tags, and we will release a `hotfix` version to fix this together as soon as possible.

Thank you for the PR!

PR #222, opened 2023-02-26
fix typos of update in question_common

bazingame:
fix typos of _update_ in question_common.go


aichy126:
Misspelling correction of words

PR #227, opened 2023-02-27
updated translation for Zh_CN and added action names for posts and tag_info memberAction

shiningnight:
the code part about memberAction localization cannot be merged util the corresponding keys propagated through all the language files, so i'm not committing it yet
but coming soon

### DIFF
--- a/i18n/zh_CN.yaml
+++ b/i18n/zh_CN.yaml
@@ -431,6 +431,9 @@ ui:
     created_at: 创建于
     edited_at: 编辑于
     history: 历史
+    action:
+      edit: 编辑
+      delete: 删除
     synonyms:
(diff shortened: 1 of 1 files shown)


shiningnight:
> 

ok, but what for the memberAction part? could i create new keys there? i don't think so

shiningnight:
> > > 
> > 
> > 
> > ok, but what for the memberAction part? could i create new keys there? i don't think so
> 
> Thanks for your feedback. You can submit a PR for the `en_US.yaml`, and then other languages will have the corresponding key. If you just add the key only for `zh_CN.yaml`, the other languages won't have it.

oh i see the correct way now, i'll close this for now

PR #239, opened 2023-03-05
I18N pt-BR.

pontovinte:


pontovinte:
> @pontovinte Thank you for your contribution to the translation, we will merge this PR in the next release version.

You're welcome. I don't know if that's the right way of sending those changes, but please, let me know if there's anything I can help. Is there any place where I can see the releases schedule?

Issue #151, opened 2023-01-03
email validation does not allow first.second@xxx.com during account setup


rockonedege:
while setting up the site,my email in the form of first.second@xxx.com was not allowed, most likely caused by the '.' in the name.

Fortunately, I could fake one and then change it to the correct one later after loggin.

shuashuai:
Thanks for your feedback, we will fix this bug in 1.0.2

fyzhu:
not yet in v1.0.4

PR #242, opened 2023-03-06
fix: email validation issued in #151

dykderrick:
Update the regex for email validation to include special characters per [RFC 5322](https://datatracker.ietf.org/doc/html/rfc5322). See Also a [Stack Overflow answer](https://stackoverflow.com/a/201378/12786353) for this one.


shuashuai:
First of all thank you very much for your PR,we are already unifying some general regular check rules，you can see from the [pattern.ts](https://github.com/answerdev/answer/blob/main/ui/src/common/pattern.ts).
image
So we use this rule to check all mailboxes in v1.0.6.If you have other questions, welcome to continue to feedback.

dykderrick:
that one looks great! I can close this PR now.

PR #397, opened 2023-06-11
fix(internal/cli/build): Using forward slash with embed fs

### LINKED ISSUE #396: Using `answer build --with <plugin>` results in `build failed open build\asset-manifest.json: file does not exist`
**Describe the bug**
I'm working on adding a plugin for DexIdP as part of issue #36, while developing it I wanted to test the plugin I've made and encountered trouble building answer with my plugin (or any other plugin like the supported github connector plugin) 

**To Reproduce**
1. Go to [Answer Releases](https://github.com/answerdev/answer/releases/tag/v1.1.0)
2. Download answer v1.1.0
3. Extract and execute the following command:
```bash
.\answer.exe build --with github.com/itamadev/plugins/connector/dexidp@v0.0.0-20230610145225-757e43bd8501 --output ./new-answer.exe
```
5. See output + error: `build failed open build\asset-manifest.json: file does not exist`

```bash
try to build a new answer with plugins:
github.com/itamadev/plugins/connector/dexidp@v0.0.0-20230610145225-757e43bd8501
[build] tmp dir: C:\code\@github.com\itamadev\answer_build3203089807
[go mod tidy]
go: finding module for package github.com/itamadev/plugins/connector/dexidp
go: finding module for package github.com/answerdev/answer/cmd
go: found github.com/answerdev/answer/cmd in github.com/answerdev/answer v1.1.0
go: found github.com/itamadev/plugins/connector/dexidp in github.com/itamadev/plugins/connector/dexidp v0.0.0-20230610145225-757e43bd8501
[go mod vendor]
try to merge i18n files
i18n dir:  C:\<path-to-build>\vendor\github.com\itamadev\plugins\connector\dexidp\i18n
try to replace ui build directory
build failed open build\asset-manifest.json: file does not exist
```

**Expected behavior**
The command should install the new plugin and result in a new binary for running answer.dev

**Platform (please complete the following information):**
 - OS: Windows 11
 - Version answer binary 1.1.0 for windows

itamadev:
This PR fixes #396, I changed the `copyDirEntries()` function implementation to use fs.WalkDir()` instead of implementing it in a loop/recursion

Machines Tested:
- [x] Windows   
- [x] Linux (on Docker)
- [ ] macOS - best if this is tested on an actual macOS machine before it's accepted

### DIFF
--- a/internal/cli/build.go
+++ b/internal/cli/build.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"embed"
 	"fmt"
+	"io"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path"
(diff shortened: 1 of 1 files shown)


itamadev:
> @itamadev Sorry, we found a flaw in the functioning of the method `fs.WalkDir()`. This function considers the softlink to be a file and not a folder. So when copying, it causes some files to not be copied correctly. We may need to change back to the original method to ensure consistent behavior.
> 
> ```go
> // WalkDir walks the file tree rooted at root, calling fn for each file or
> // directory in the tree, including root.
> //
> // All errors that arise visiting files and directories are filtered by fn:
> // see the fs.WalkDirFunc documentation for details.
> //
> // The files are walked in lexical order, which makes the output deterministic
> // but requires WalkDir to read an entire directory into memory before proceeding
> // to walk that directory.
> //
> // WalkDir does not follow symbolic links found in directories,
> // but if root itself is a symbolic link, its target will be walked.
> func WalkDir(fsys FS, root string, fn WalkDirFunc) error {
> 	info, err := Stat(fsys, root)
> 	if err != nil {
> 		err = fn(root, nil, err)
> 	} else {
> 		err = walkDir(fsys, root, &statDirEntry{info}, fn)
> 	}
> 	if err == SkipDir || err == SkipAll {
> 		return nil
> 	}
> 	return err
> }
> 
> ```
> **WalkDir does not follow symbolic links found in directories**
> 
> ### test code
> ```go
> 	dir := "****/ui/node_modules/.pnpm/@ampproject+remapping@2.2.0/node_modules/@jridgewell/"
> 	fs.WalkDir(os.DirFS(dir), ".", func(path string, d fs.DirEntry, err error) error {
> 		fmt.Println(d.Name(), "is dir:", d.IsDir())
> 		return nil
> 	})
> 
> 	stat, err := os.Stat(dir + "gen-mapping")
> 	if err != nil {
> 		fmt.Println(err)
> 	} else {
> 		fmt.Println("gen-mapping is dir:", stat.IsDir())
> 	}
> 	return
> ```
> 
> the output
> ```
> gen-mapping is dir: false
> 
> gen-mapping is dir: true
> ```
> 
> `os.Stat` and `fs.WalkDir` behavior is different.
> 
> 
> #### ls command output
> ```
> ls -al
> total 0
> drwxr-xr-x  4 linkinstar  staff  128 Jul 12 15:07 .
> drwxr-xr-x  4 linkinstar  staff  128 Jul 12 15:07 ..
> lrwxr-xr-x  1 linkinstar  staff   75 Jul 12 15:07 gen-mapping -> ../../../@jridgewell+gen-mapping@0.1.1/node_modules/@jridgewell/gen-mapping
> lrwxr-xr-x  1 linkinstar  staff   80 Jul 12 15:07 trace-mapping -> ../../../@jridgewell+trace-mapping@0.3.15/node_modules/@jridgewell/trace-mapping
> ```

@LinkinStars Well I understand the problem, sorry for not testing it thoroughly enough.

I think I can implement a fix while using the original method of walking down the directory, so the issue we fixed using the `fs.WalkDir()` method won't resurface, 
I'll revert the changes and see if I find a fix and update on it.

PR #405, opened 2023-06-16
fix: Truncate long username with the ellipsis

### LINKED ISSUE #404: Very long username breaks though the container on Users page
**Describe the bug**
On the "Users" page, when a user has a very long username, it will makes his username break though the container and overlapped with the profile picture of the next user. 

**To Reproduce**
1. Set a very long username
2. Go to 'Users'
3. See error

**Expected behavior**
Username needs to be truncated with the ellipsis.

**Platform**
 - Device: Desktop
 - OS: macOS
 - Browser: Chrome
 - Version: 114

conr2d:
Fix #404

### DIFF
--- a/ui/src/index.scss
+++ b/ui/src/index.scss
@@ -132,6 +132,12 @@ a {
   }
 }
 
+.text-ellipsis {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
 .text-truncate-1,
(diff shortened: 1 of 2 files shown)


haitaoo:
As described here: [#404](https://github.com/answerdev/answer/issues/404#issuecomment-1594171680), This will be fixed in other ways in the next version. Thanks a lot!

Issue #423, opened 2023-07-01
Display questions in the question content preview on the Edit Answer page


zhoushengdao:
**Describe the bug**

*I am not sure if this issue only exists on `meta.segmentfault.com`.*

Strange graphics appeared in the preview of the question content, as shown in the following picture.

**To Reproduce**
1. Go to <https://meta.segmentfault.com/questions/10010000053631176>
2. Scroll down to the "Answers" section
3. Click on "Edit" button of an answer

**Expected behavior**
This should not be displayed.

**Screenshots**
![image](https://github.com/answerdev/answer/assets/53334669/285f8748-cbf0-43de-8520-be8cd8510259)
▲ on Microsoft Edge 114.0.1823.67

![image](https://github.com/answerdev/answer/assets/53334669/ab4ebf5c-8340-422c-a89b-85ebac0b0dc3)
▲ on Google Chrome 117.0.5863.0 (canary)

![image](https://github.com/answerdev/answer/assets/53334669/ebdc0c64-6e88-422b-ba6f-67339b7a2796)
▲ on Firefox Developer Edition 115.0b9 (Please zoom in, there are still some issues)

**Platform (please complete the following information):**
 - Device: Desktop
 - OS: Windows 11
 - Browser: Edge 
 - Version: 114.0.1823.67

PR #428, opened 2023-07-04
fix: answer edit style

MMmaXingXing:
1. I think this editor maybe change 
image
image
Reproduction steps: Answer the question, click the edit button to enter the editing interface, and then display.

2. create user max display_name is 4 - 30 but message is 3 -30

### DIFF
--- a/ui/src/pages/Questions/EditAnswer/index.scss
+++ b/ui/src/pages/Questions/EditAnswer/index.scss
@@ -8,7 +8,6 @@
       height: calc(100% - 12px);
       overflow-y: scroll;
       z-index: 1;
-      background-color: #eee;
     }
     .scroll-bar {
       height: 12px;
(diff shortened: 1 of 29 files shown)


haitaoo (approved):
Thank you for the **PR**, `display_name` request description did not have a more code-synchronised update before.

ui/src/pages/Questions/EditAnswer/index.scss
```diff
@@ -8,6 +8,7 @@
       height: calc(100% - 12px);
       overflow-y: scroll;
       z-index: 1;
+      background-color: #eee;
```
haitaoo, line 11: This change is not particularly necessary. Also not particularly appropriate - the `.content` node has a `.bg-white` className, which would have to be cleaned up first if changed.

Suggest cancelling the changes to this file and I'll merge #428.

Thanks!

haitaoo (changes requested):

ui/src/pages/Questions/EditAnswer/index.scss
zhoushengdao: This is currently being fixed in #423.

PR #500, opened 2023-08-29
feat(search): Internal search to be case insensitive

heubeck:
The internal search currently is case sensitive.
There are properitary options on each supported database (sqlite, mysqsl, postgres) but they have none in common.
The `UPPER` function is available for all of them.

As I'm not firm with xorm, please have a close look ;)

### DIFF
--- a/internal/repo/search_common/search_repo.go
+++ b/internal/repo/search_common/search_repo.go
@@ -107,12 +107,12 @@ func (sr *searchRepo) SearchContents(ctx context.Context, words []string, tagIDs
 	likeConQ := builder.NewCond()
 	likeConA := builder.NewCond()
 	for _, word := range words {
-		likeConQ = likeConQ.Or(builder.Like{"title", word}).
-			Or(builder.Like{"original_text", word})
+		likeConQ = likeConQ.Or(builder.Expr("UPPER(title) LIKE UPPER(?)", "%"+word+"%")).
+			Or(builder.Expr("UPPER(original_text) LIKE UPPER(?)", "%"+word+"%"))
 		argsQ = append(argsQ, "%"+word+"%")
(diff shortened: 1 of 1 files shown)


heubeck:
ah, ok - found only table schema related stuff like COLLATE.

thx - will try to reconfigure my postgres.

heubeck:
Ok, crawled the internet and it seems, that there's not case-insensitive global charset or table  collation for postgres that supports `like` queries. one have to work with a specific `CITEXT` datatype, or using `ilike` operator.

WDYT is the less worse option? implementing a postgres dialect in the search_repo, having a different schema for postgres (using citext, but probably also different indexes), something else?

In general, I'd like to use an external search engine (MeiliSearch) - but because of https://github.com/answerdev/plugins/issues/23 and other not yet reported issues (searches with options don't work properly, lots of not yet analyzed http 400 in the MeiliSearch log), that's not an option atm.

heubeck:
> Since the MeiliSearch plugin is still in beta, it does have a lot of issues. If you have any problems, you can submit your issue, no worries at all. We will arrange to upgrade and solve them in the following versions.

Thank you @LinkinStars, will try to test and report everything around MeiliSearch, an opportunity to (manually) build a full search index after switching on the plugin would help, so maybe that's the first issue to address.

So let's close this PR, was at least a good way for me to get into answer. til next time ;)

heubeck:
> MeiliSearch plugin has been updated. Existing data will be synchronised to MeiliSearch when the MeiliSearch configuration is re-saved.

Awesome, thanks @LinkinStars.
Is it already contained in the all-in-one image (was updated 2h ago)?

heubeck:
Yehaa ;)
Just activated it, thx. Will report any finding to the plugin repo.

PR #531, opened 2023-09-19
fix(image): support upload jog images

buttercannfly:
Currently, answer don't support upload JPG image, it will be blocked at decode stage


buttercannfly:
before change the IsSupportedImageFile function, it will fail
Screenshot2023_09_19_153552
after changing, it successfully uploaded
image

Issue #532, opened 2023-09-19
Can't handle jpg images


buttercannfly:
**Describe the bug**
Can't successfully handle JPG file upload

**To Reproduce**
1. Go to 'Post Question' or 'Answer Question'
2. Click on '.Upload image'
3. Upload jpg image
4. See error “Unsupported file format.”

**Expected behavior**
I have reviewed the code and I found that backend try to use "image/jpeg" to support jpg type images

**Screenshots**
<img width="545" alt="image" src="https://github.com/answerdev/answer/assets/36593447/44e96186-5fc0-4b29-a496-eef0a67543ef">

example jpg image
![run-local-env 0baf8195](https://github.com/answerdev/answer/assets/36593447/bf2967e8-73a7-47da-9f52-2854bd4621e2)

buttercannfly:
https://github.com/answerdev/answer/pull/533

wangxue1990:
The problem still exists

shuashuai:
> The problem still exists

Which version of Answer are you using? You can go to the [official website](https://meta.answer.dev/ ) to test whether the problem still exists, if so, please provide us with the detailed steps and error messages.

wangxue1990:
> > The problem still exists
> 
> Which version of Answer are you using? You can go to the [official website](https://meta.answer.dev/) to test whether the problem still exists, if so, please provide us with the detailed steps and error messages.
Thank you.  I am using the latest version compiled by you; This link was tested using the system you deployed, and the uploaded JPG cannot be displayed; If you check the upload folder on the server, you should be able to find the corresponding jpg file, but the size is 0 bytes;
test for upload jpg https://meta.answer.dev/questions/D1RR2?shareUserId=wangxue

shuashuai:
> > > The problem still exists
> > 
> > 
> > Which version of Answer are you using? You can go to the [official website](https://meta.answer.dev/) to test whether the problem still exists, if so, please provide us with the detailed steps and error messages.
> > Thank you.  I am using the latest version compiled by you; This link was tested using the system you deployed, and the uploaded JPG cannot be displayed; If you check the upload folder on the server, you should be able to find the corresponding jpg file, but the size is 0 bytes;
> > test for upload jpg https://meta.answer.dev/questions/D1RR2?shareUserId=wangxue

I tested that both jpg and jpeg can be uploaded normally. You can change a few pictures and try it. You can also test it on other communities to see if there are any restrictions on this machine.

[...]

PR #533, opened 2023-09-19
better way to decode images

buttercannfly:
In the past, answer use postfix of file to choose decoder
now, we init the decoder of image, and use image.decode, it will ignore the postfix


buttercannfly:
Thanks, maybe I can delete the support for gif. I just saw there published image/gif.

buttercannfly:
close #532

Issue #590, opened 2023-10-27
License breaks helm chart


mortenbirkelund:
**Describe the bug**
After adding the license header in https://github.com/apache/incubator-answer/commit/3d2187d8491d582a41481e8a52f142e2750e500e
the helm chart doesn work anymore, because the apiversion of the serviceaccount is being mistakenly seen as a comment. 

Removing the license header, and then the chart works again. This only happens in the service account, because it is the only one that is deployed conditional depending on the values. 

https://github.com/apache/incubator-answer/blob/main/charts/templates/serviceaccount.yaml

See an explanation here: 
https://github.com/argoproj/argo-helm/issues/507#issuecomment-793127686

**To Reproduce**
1. Go to the charts directory
2. run ```helm template .```

**Expected behavior**
The serviceaccount rendered correctly
![image](https://github.com/apache/incubator-answer/assets/8955039/7f03975a-83f8-4d9b-8406-eaa0009cb874)

**Screenshots**
The serviceaccount is missing the apiversion, because it is seen as a comment
![image](https://github.com/apache/incubator-answer/assets/8955039/ef633c6b-1a43-4564-ac4d-c6c8a0fb2e27)

**Platform (please complete the following information):**
 - Device: [e.g. Desktop, Mobile]
 - OS: [e.g. macOS]

mortenbirkelund:
Explanation can be seen here
https://github.com/helm/helm/issues/4191

Can be solved by changing
```{{- if .Values.serviceAccount.create -}}```

to
```{{ if .Values.serviceAccount.create -}}```

PR #591, opened 2023-10-28
fix: License breaks helm chart

sy-records:
fix apache/incubator-answer\#590


mortenbirkelund:
I was mistaken. The same thing goes for the hpa, pvc and ingress. They are also conditional are are giving the same errors

LinuxSuRen (commented):
I'm wondering if it's possible to have a way to test aginst it automatically. For example, doing a lint or a dry run something.

LinuxSuRen:
By the way, test the Helm chart via e2e testing might be a good choice. During the e2e testing, it can install the Helm in k3s or other small Kubernetes distribution.

tisonkun (approved):
Thank you!

PR #611, opened 2023-11-04
refactor: remove redundant `len` and `nil` check

Juneezee:
From the Go specification (https://go.dev/ref/spec#For_range):

> "1. For a nil slice, the number of iterations is 0."

`len` returns 0 if the slice is nil (https://pkg.go.dev/builtin#len). Therefore, checking `len(v) > 0` around a loop is unnecessary.

Example Go playground: https://go.dev/play/p/s3FZcKC8ltm


Juneezee (commented):
Tip: Hide whitespace changes to make the diff easier to read

image

image

PR #614, opened 2023-11-07
fix: issue-609

### LINKED ISSUE #609: The description of first post in Revision select is unclear
**To Reproduce**

Edit question/answer/tag:
1. Select 'revision'
2. The first revision's description is "Edit xxx" 

**Expected behavior**

The first revision's description is "Add xxx"
- Add question
- Add answer
- Add tag

**Screenshots**

image

kelvinkuo:
fix #609

### DIFF
--- a/i18n/nl_NL.yaml
+++ b/i18n/nl_NL.yaml
@@ -463,8 +463,7 @@ ui:
   edit_tag:
     title: Edit Tag
     default_reason: Edit tag
-    default_first_reason: Add tag
-    default_first_reason: Add tag    form:
+    form:
       fields:
         revision:
           label: Revision
@@ -508,7 +507,6 @@ ui:
   edit_answer:
     title: Edit Answer
     default_reason: Edit answer
-    default_first_reason: Add answer
     form:
       fields:
         revision:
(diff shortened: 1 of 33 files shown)


i18n/nl_NL.yaml
```diff
@@ -463,7 +463,8 @@ ui:
   edit_tag:
     title: Edit Tag
     default_reason: Edit tag
-    form:
+    default_first_reason: Add tag
+    default_first_reason: Add tag    form:
```
shuashuai, line 467: It’s repeated here. You need to modify it. Also pay attention to the format.

shuashuai (changes requested):
First of all, thank you very much for your contribution. What you need to know about i18n files and how to contribute.

Our translation uses [crowdin](https://crowdin.com/project/answer), so if you contribute by changing the language file in the code, then you only need to change en_US.yaml, other files will be overwritten after they are synchronized on crowdin.

kelvinkuo:
@shuashuai Thanks for your expaination, please review one more time.

shuashuai (approved):
That's great,we will merge this PR as soon as possible.

PR #618, opened 2023-11-10
support tag search for synonyms

### LINKED ISSUE #587: tag search need support for synonyms
**Describe the bug**

https://meta.answer.dev/tags

I set "k8s" as a synonym for "kubernetes". When I search "k8s" in `/tags` page, there is no results.

**To Reproduce**

1. Go to `/tags`
2. Click on "filter by tag name", type `k8s`
3. See no result

**Expected behavior**

Return tag "kubernetes", like search tags in "ask question" page

kelvinkuo:
fix #587 

This commit adds 1 or 2 db query for synonyms tag.

- no more query, when the slugName == ""
- 1 more query if slugName not matches any tag or slugName is not a synonyms
- 2 more query if slugName is just a synonyms

I think the cost is cheep slugName has index.

Off topic, the design of synonyms is a bit weird,
- A normal tag disspears when it became a synonyms tag
- They are not shown in tag list（So I did not know the 'synonyms' thing until I saw the issue-587.）
- It's so hidden, not very clear to users.

### DIFF
--- a/internal/repo/tag_common/tag_common_repo.go
+++ b/internal/repo/tag_common/tag_common_repo.go
@@ -22,6 +22,7 @@ package tag_common
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/apache/incubator-answer/internal/base/data"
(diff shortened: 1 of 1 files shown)


kelvinkuo:
> When `main_tag_id != 0`(means this tag is a synonym) and `slug_name` is equal the query condition, the query result maybe what we want. How about that?

The query result is synonym and you need another query to get main tag info. Right?

kelvinkuo:
> I think this is a bit complicated, maybe you just need to adjust the query `session.Where("main_tag_id = 0")`. When `main_tag_id = 0` means this tag is main tag not a synonym. When `main_tag_id != 0`(means this tag is a synonym) and `slug_name` is equal the query condition, the query result maybe what we want. How about that?

Please review one more time, Is this what you mean?

@LinkinStars

PR #621, opened 2023-11-11
fix #587

### LINKED ISSUE #587: tag search need support for synonyms

shubo83:
1. query for synonyms and get the corresponding mainTagIds.
2. add ID in condition to the main query statement.

### DIFF
--- a/internal/repo/tag_common/tag_common_repo.go
+++ b/internal/repo/tag_common/tag_common_repo.go
@@ -167,10 +167,17 @@ func (tr *tagCommonRepo) GetTagPage(ctx context.Context, page, pageSize int, tag
 	tagList []*entity.Tag, total int64, err error,
 ) {
 	tagList = make([]*entity.Tag, 0)
-	session := tr.data.DB.Context(ctx)
 
+	var mainTagIds []string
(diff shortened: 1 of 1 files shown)


shubo83:
> I noticed you changed the code, but there are still some issues. You can test it by following steps.
> 
> 1. Add two tags `a` and `b`
> 2. Set `b` to be a synonym for `a`
> 3. Use `b` for querying via api
> 
> You'll notice that the `a` tag still doesn't appear in the query result. It doesn't solve the issue of #587 .

ok, I'll try to debug it.

shubo83:
> I noticed you changed the code, but there are still some issues. You can test it by following steps.
> 
> 1. Add two tags `a` and `b`
> 2. Set `b` to be a synonym for `a`
> 3. Use `b` for querying via api
> 
> You'll notice that the `a` tag still doesn't appear in the query result. It doesn't solve the issue of #587 .

I followed your three steps to test, and there was no problem. As shown below:

image

image

Strange, where is the problem?

PR #622, opened 2023-11-14
fix(i18n): fix time format in de-DE

xialeistudio:
Fix time format problem in **de-DE**


shuashuai (approved):
Thanks for your contribution, we will merge it as soon as possible.

PR #664, opened 2023-11-28
fix: can't show tag description in tag edit page

### LINKED ISSUE #637: Can't show tag description in tag edit page
**To Reproduce**
1. Enter a tag info page with description
2. Click on 'Edit' enter edit page
3. The description show nothing

**Expected behavior**
Show previously filled description.

**Screenshots**

image

hgaol:
fix #637 .

### DIFF
--- a/ui/src/pages/Tags/Edit/index.tsx
+++ b/ui/src/pages/Tags/Edit/index.tsx
@@ -83,25 +83,11 @@ const Index = () => {
   });
 
   useEffect(() => {
-    setFormData({
-      ...formData,
-      displayName: { ...formData.displayName, value: data?.display_name || '' },
-      slugName: { ...formData.slugName, value: data?.slug_name || '' },
-      description: {
-        ...formData.description,
-        value: data?.original_text || '',
-      },
-    });
-
-    setImmData({
-      ...immData,
-      displayName: { ...immData.displayName, value: data?.display_name || '' },
-      slugName: { ...immData.slugName, value: data?.slug_name || '' },
-      description: {
-        ...immData.description,
-        value: data?.original_text || '',
-      },
-    });
+    initFormData.displayName.value = data?.display_name || '';
+    initFormData.slugName.value = data?.slug_name || '';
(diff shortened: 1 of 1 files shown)


ui/src/pages/Tags/Edit/index.tsx
```diff
@@ -85,6 +82,18 @@ const Index = () => {
...
+  useEffect(() => {
+    setFormData({
+      ...formData,
+      displayName: { ...formData.displayName, value: data?.display_name || '' },
+      slugName: { ...formData.slugName, value: data?.slug_name || '' },
+      description: {
+        ...formData.description,
+        value: data?.original_text || '',
+      },
+    });
```
shuashuai, line 94: You may have overlooked immData. immData is used to record the initial value. When the user changes the content and leaves the page, there will be a prompt; therefore, you need to add an assignment operation to immData here;

hgaol: Thanks @shuashuai , updated. There's one thing makes me confused that for below codes.
```
    useEffect(() => {
    ....
    if (!display_name || !slug_name || !original_text) {
      return;
    }
    ....
    }
```
Seems it'll never go into this condition branch.

shuashuai: > Thanks @shuashuai , updated. There's one thing makes me confused that for below codes.
> 
> ```
>     useEffect(() => {
>     ....
>     if (!display_name || !slug_name || !original_text) {
>       return;
>     }
>     ....
>     }
> ```
> 
> Seems it'll never go into this condition branch.

Yes, you can remove this condition now! In addition, for the previous assignment operation, you can simplify the code and use an object to assign the value. Try to reduce the number of lines of code! Thank you for your thoughts and suggestions！

hgaol: resolved. please help to review and let me know if any comment. thanks!

shuashuai (approved):

Issue #661, opened 2023-11-28
Disable unlisted post from showing up in profile and search engines


fenbox:
## Currently

- Unlisted posts still shows in user profile.
- Unlisted posts can be indexed by search engine.

## Expectation

- Hide unlisted posts in user profile from other users.
- Post author, Admin, Moderator still can see.
- Block Search indexing with [`noindex`](https://developers.google.com/search/docs/crawling-indexing/block-indexing).

insistedMarin:
Hello, i am new here, and i would like try to solve this problem. In fact, I have partially fixed the display problem of the user profile. But for the search problem of unlisted posts, it worked fine in my tests. Has this problem been fixed? Also, please consider assigning this task to me.

PR #672, opened 2023-12-10
Fix/1.2.0/userprofile (#661)

insistedMarin:
This Pull Request introduces a new filtering functionality to the `userprofile` page. The main objective of this update is to enhance the user experience by providing a more refined control over the visibility of posts.

**Key Changes:**
1. **Filter Implementation for `topData` and `listData`:** The `topData` and `listData` on the `userprofile` page have been updated with a filtering mechanism. This feature filters out 'unlisted' posts, ensuring that they are only visible to specific user roles.

2. **Role-Based Visibility:** The visibility of 'unlisted' posts is now restricted. These posts are only displayed to the Post author, Admin, and Moderator roles. 

**Testing:**
The new filter functionality has been thoroughly tested to ensure it operates as intended across various scenarios. Tests included checking the visibility of posts for different user roles and verifying the correct functioning of the `topData` and `listData` filters.

Thank you for considering this Pull Request.

Related issue: #661

### DIFF
--- a/ui/src/pages/Users/Personal/index.tsx
+++ b/ui/src/pages/Users/Personal/index.tsx
@@ -27,6 +27,7 @@ import {
   usePersonalTop,
   usePersonalListByTabName,
   questionDetail,
+  getAnswers,
 } from '@/services';
 import { usePageTags } from '@/hooks';
 import { Pagination, FormatTime, Empty } from '@/components';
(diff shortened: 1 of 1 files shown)


shuashuai:
First of all thank you for your contribution!

A good solution should be to adjust the results returned by the corresponding API through token authentication! Calling the details API in batches through the list is not a good solution! 

As developers, we should think more about whether the solution is reasonable! This is a very important concept!

insistedMarin:
Thank you very much for your guidance, I will revisit this issue !

shuashuai:
> Thank you very much for your guidance, I will revisit this issue !

This requires adjusting some logic on the backend. If you need technical support, you can contact @LinkinStars  for help.

hgaol:
if needed, I can help with the backend part. :)

insistedMarin:
> if needed, I can help with the backend part. :)

Many thanks if you can provide any help.

PR #685, opened 2023-12-15
fix: questions with synonym tags do not show under the origin tag

### LINKED ISSUE #680: After setting the tag as a synonym, the questions do not show under the origin tag.
**To Reproduce**

1. Two questions:
   - Question A with tag "k8s"
   - Question B with tag "kubernetes"
2. Set "k8s" as a synonym for "kubernetes".
3. Question A do not show under "kubernetes" tag.

**Expected behavior**

Question A show under "kubernetes" tag.

hgaol:
fix #680 .

### DIFF
--- a/internal/repo/question/question_repo.go
+++ b/internal/repo/question/question_repo.go
@@ -23,18 +23,13 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/apache/incubator-answer/plugin"
-	"github.com/segmentfault/pacman/log"
 	"strings"
 	"time"
 	"unicode"
-	"xorm.io/xorm"
-
-	"github.com/apache/incubator-answer/internal/base/handler"
-	"xorm.io/builder"
 
 	"github.com/apache/incubator-answer/internal/base/constant"
 	"github.com/apache/incubator-answer/internal/base/data"
+	"github.com/apache/incubator-answer/internal/base/handler"
 	"github.com/apache/incubator-answer/internal/base/pager"
 	"github.com/apache/incubator-answer/internal/base/reason"
 	"github.com/apache/incubator-answer/internal/entity"
@@ -43,6 +38,10 @@ import (
 	"github.com/apache/incubator-answer/internal/service/unique"
 	"github.com/apache/incubator-answer/pkg/htmltext"
 	"github.com/apache/incubator-answer/pkg/uid"
+	"github.com/apache/incubator-answer/plugin"
+	"github.com/segmentfault/pacman/log"
+	"xorm.io/builder"
+	"xorm.io/xorm"
 
 	"github.com/segmentfault/pacman/errors"
 )
@@ -347,15 +346,15 @@ func (qr *questionRepo) SitemapQuestions(ctx context.Context, page, pageSize int
 }
 
 // GetQuestionPage query question page
-func (qr *questionRepo) GetQuestionPage(ctx context.Context, page, pageSize int, userID, tagID, orderCond string, inDays int) (
+func (qr *questionRepo) GetQuestionPage(ctx context.Context, page, pageSize int, tagIDs []string, userID, orderCond string, inDays int) (
 	questionList []*entity.Question, total int64, err error) {
 	questionList = make([]*entity.Question, 0)
 
 	session := qr.data.DB.Context(ctx).Where("question.status = ? OR question.status = ?",
 		entity.QuestionStatusAvailable, entity.QuestionStatusClosed)
-	if len(tagID) > 0 {
+	if len(tagIDs) > 0 {
 		session.Join("LEFT", "tag_rel", "question.id = tag_rel.object_id")
-		session.And("tag_rel.tag_id = ?", tagID)
+		session.In("tag_rel.tag_id", tagIDs)
 		session.And("tag_rel.status = ?", entity.TagRelStatusAvailable)
 	}
 	if len(userID) > 0 {
--- a/internal/repo/search_common/search_repo.go
+++ b/internal/repo/search_common/search_repo.go
@@ -98,7 +98,7 @@ func NewSearchRepo(
 }
 
 // SearchContents search question and answer data
-func (sr *searchRepo) SearchContents(ctx context.Context, words []string, tagIDs []string, userID string, votes int, page, size int, order string) (resp []*schema.SearchResult, total int64, err error) {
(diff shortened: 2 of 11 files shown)


hgaol:
> @hgaol Firstly, thank you very much for your contribution.
> 
> 
> 
> In the page `http://127.0.0.1:8080/tags/kubernetes`, user can find all questions that search by synonym tag. That's right you've fixed the issue in this part.
> 
> 
> 
> But in the page `http://127.0.0.1:8080/search?q=%5Bkubernetes%5D`, user can search content via expression such as `[tag]`. This place still has the same issue that hasn't been fixed.

Thanks for figuring it out! I'll fix it later.

kumfo:
> update,
> 
> after dive into the SQL logic, I guess the original design is option1. So for the synonym scenario, I think it may need a 2D array for tags. Each dimension is for each tag and its mainTag and synonym tags. Takes 2 tags as example, the logic could be
> 
> ```
> AND tag_rel0.status = ?
> AND tag_rel0.tag_id IN (?, ?)
> AND tag_rel1.status = ?
> AND tag_rel1.tag_id IN (?)
> ```
> 
> Hi @LinkinStars , one quick question, what's the expected logic when there're multiple tags? option 1 or 2? I think for synonym tags, it's option 2. what about if 2 different tags?
> 
> option 1: return objects with both `tag1` and `tag2`. option 2: return objects with either `tag1` or `tag2`.

Currently, SQL supports multi label search. If you need to support synonym label search, you should directly query synonym labels when parsing the underlying labels. This way, future search plugins can also be used.

The Search Parse Code ：

[search parse](https://github.com/apache/incubator-answer/blob/68f5af01d9d57fbfe7b7920dd91ff4f62029ac59/internal/service/search_parser/search_parser.go#L108)

hgaol:
Hi @kumfo , I've tested adding `ts.tagRepo.GetIDsByMainTagId` and adding below lines
```
synIDs, err := sp.tagCommonService.GetTagIDsByMainTagID(ctx, tag.ID)
...
tagIDs = append(tagIDs, synIDs...)
```

but found it'll only return the objects with all these tags. You can find the SQL will query tags with below commands. IMO, this is not solve the issue. Take `k8s` and `kubernates` as example, when I search [k8s], the expected objects are which have either `k8s` or `kubernates` tag, right?
```
...
      `post_update_time`
    FROM
      `question`
      INNER JOIN tag_rel as tag_rel0 ON question.id = tag_rel0.object_id
      INNER JOIN tag_rel as tag_rel1 ON question.id = tag_rel1.object_id
    WHERE
      `question`.`status` < 10
      AND `question`.`show` = 1
      AND tag_rel0.status = 1
      AND tag_rel0.tag_id = tagId1
      AND tag_rel1.status = 1
      AND tag_rel1.tag_id = tagId2
...
```

kumfo:
> scenario

Sorry! I know what the problem is, and I overlooked the previous search logic. The original tag search must include all these tags at the same time. If you want to modify it now, this search logic will probably need to be processed during the search query.

You need to change the code in file `[internal/repo/search_common/search_repo.go](https://github.com/apache/incubator-answer/blob/152934c740d163c2b6345ac2e8c8f542957efa75/internal/repo/search_common/search_repo.go#L150)`, and you search the comments of `// check tag` in file. you add synonymous tags query and add `or` conditions to SQL.

However, if possible, you can parse the synonymous tags in `search_parse.go`, obtain a separate query parameter, and submit it to this search query. Then the subsequent search `plugins` can also be modified to be compatible with this query method. but the program will  will become more complex.

hgaol:
Hi @kumfo , exactly! I've pushed new commit about the changed logic. In a word, it changes the original SQL from
```sql
AND tag_rel0.status = 1
AND tag_rel0.tag_id = tagId1
AND tag_rel1.status = 1
AND tag_rel1.tag_id = tagId2
```
to
```sql
AND tag_rel0.status = ?
AND tag_rel0.tag_id IN (?, ?)
AND tag_rel1.status = ?
AND tag_rel1.tag_id IN (?)
```

As you mentioned, it changes the interface and all the search plugin need to be modified to be compatible. I'm not sure if there's any better solution. Feel free to let me know and I can implement in this PR. Thanks!

kumfo:
@hgaol Good Job! but I saw your commit that find you not change the query conditions to `in` query, You need to change the query conditions,

Original is :
```
And(builder.Eq{
    ast + ".tag_id": tagID,
    ast + ".status": entity.TagRelStatusAvailable,
})
```

it need to change like this:
```
And(builder.In(ast+".tag_id", tagID)).
And(builder.Eq{
    ast + ".status": entity.TagRelStatusAvailable,
})
```

you can search the code comments of `// check tag`, there are 3 areas that need to be modified.

hgaol:
Hi @kumfo , thanks for your review! I found that when there's array in `builder.Eq`, it'll change it into `in`. And I've tested it. related documents.
image
https://pkg.go.dev/xorm.io/builder@v0.3.12

hgaol:
I haven't updated for `SearchQuestions` and `SearchAnswers` yet. Just want to make sure if it's the expected solution. Do I need to update `SearchQuestions` and `SearchAnswers` as well?

kumfo:
> I haven't updated for `SearchQuestions` and `SearchAnswers` yet. Just want to make sure if it's the expected solution. Do I need to update `SearchQuestions` and `SearchAnswers` as well?

Yeah, you need to update SearchQuestions and SearchAnswers as well. Because there are three ways to search, the default search includes questions and answers. If questions(use `is:question`) or answers (use `is:answer` are specified, tag searches can also be used at the same time.

hgaol:
hi @kumfo , updated. Since `Eq{ast+."tag_id": []string{"c", "d"}}` works, I didn't change it into `builder.In(ast+".tag_id", tagID)`. And I've tested that it works well.
Feel free to let me know if you prefer `builder.In(ast+".tag_id", tagID)`, and I can update and test it.

kumfo:
> hi @kumfo , updated. Since `Eq{ast+."tag_id": []string{"c", "d"}}` works, I didn't change it into `builder.In(ast+".tag_id", tagID)`. And I've tested that it works well. Feel free to let me know if you prefer `builder.In(ast+".tag_id", tagID)`, and I can update and test it.

There is no problem in handling it this way. I personally think that using the `in` explicit identifier will make it clearer for people who read the code.

hgaol:
Updated to replace with `builder.In`. For search plugins, I'm not quite familiar with and may take some time to ramp up. If it's not urgent, I'd like to try to update them to be compatible as well.

kumfo:
Hi @hgaol 

Well done!, I merged it to `dev` branch, your submissions will be released in version 1.2.5.

PR #687, opened 2023-12-17
Fix/1.2.1/userprofile backend (# 661)

insistedMarin:
This update fixes an issue where unlisted posts could still be seen by all users in a profile. Changes were made to the backend to ensure that unlisted posts are hidden from general users but remain visible to post authors, administrators, and moderators, in compliance with our privacy and content management standards.

**Main changes:**

The data call API on the userprofile page adds the LoginUserRole parameter to implement filtering functions in back-end question_service, comment_service and other services.

**test:**

The new filter functionality has been thoroughly tested to ensure it operates as expected in a variety of situations. Testing included checking the visibility of posts for different user roles.

Thank you for considering this pull request.


insistedMarin:
In addition, on the userprofile page, in addition to "question", "answer", there are multiple tables such as "collection", "comment" and "rank". Do we need to implement the same filtering function for these tables?

PR #690, opened 2023-12-21
Sanitize input from contributor fork workflows.

axisofentropy:
This update sanitizes user input from contributor forks. This fixes a GHA workflow security vulnerability.


axisofentropy:
> @axisofentropy Thank you. We had a issue earlier and wondered if you could help us out. As you can see https://github.com/apache/incubator-answer/actions/runs/7246371114 For this reason, we can't use `reusable.yaml`. So, I would like to ask if there is another way.

You can copy it into your repo, that should be straightforward.

PR #693, opened 2023-12-21
fix: i18n text

### LINKED ISSUE #692: Tags with special characters
**Describe the bug**
Special characters in tags causes answer to assume incorrect urls. For instance a forward slash will be accepted in the tag, which obviously causes a problem in a url

**To Reproduce**
1. Create a tag with a forward slash "Something / something"
2. See that the url that answer uses for the tag is https://yourdomain/tags/Something-/-something
3. Try and visit this url, and you will get a 404

**Expected behavior**
I expected that answer will either not allow special characters when creating tag, or would url-encode before creating the url. 

**Screenshots**
NIL

sy-records:
close #692 

There are several languages that require specialized translators. 😅

### DIFF
--- a/i18n/af_ZA.yaml
+++ b/i18n/af_ZA.yaml
@@ -433,7 +433,7 @@ ui:
             range: Display name up to 35 characters.
         slug_name:
           label: URL Slug
-          desc: 'Must use the character set "a-z", "0-9", "+ # - ."'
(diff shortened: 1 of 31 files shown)


shuashuai (approved):
Thanks for your contribution!  We will merge this pr ASAP.

PR #699, opened 2023-12-27
ci: use apache-answer- as release artifact name

tisonkun:


.goreleaser.yaml
```diff
@@ -27,6 +27,7 @@ before:
 
 release:
   draft: true
+  name_template: apache-answer-{{ .Version }}-incubating-src
```
tisonkun, line 30: According to https://goreleaser.com/customization/archive/, this cause the extracted folder named as "apache-answer-x.y.x-incubating-src"

tisonkun:
@LinkinStars @joyqi where do we get the "apache-answer-x.y.x-incubating-src" tarball now? I'd like to change its extracted folder to follow the same name instead of "incubating-answer-{tag}`.

tisonkun:
Seems not with Goreleaser but some manual steps that are missing on the release process.

Let's document it and perhaps let's use `git-archive` to create the source tarball which is reproducible and programmable.

PR #701, opened 2023-12-29
fix issue:userprofile hide unlisted post(#661)

insistedMarin:
This update fixes a display issue with the question list in profiles

Main changes:

Change the query conditions of the "GetQuestionPage" method of "question_repo" and decide whether to obtain hidden posts from the database based on the new parameter "showHidden".

The "showHidden" parameter is created in the "GetQuestionPage" method of "question_service" based on the role of the logged in user.

test:

The new filter functionality has been thoroughly tested to ensure it operates as expected in a variety of situations. Testing included checking the visibility of posts for different user roles.

Thank you for considering this pull request.

### DIFF
--- a/internal/repo/question/question_repo.go
+++ b/internal/repo/question/question_repo.go
@@ -347,7 +347,7 @@ func (qr *questionRepo) SitemapQuestions(ctx context.Context, page, pageSize int
 }
 
 // GetQuestionPage query question page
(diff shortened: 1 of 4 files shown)


insistedMarin:
Thank you very much for your guidance, I opened a new PR to solve this issue from the dev branch.
image
The project compiles correctly, I think the reason it didn't compile before was that I forgot to upload a few auto-generated files ?

PR #715, opened 2024-01-09
fix: the answer's content in edit answer page disappears

### LINKED ISSUE #713: In some cases, the answer's content in "edit answer" page disappears
**To Reproduce**

1. On the question page, click "Edit" under a specific answer.
2. Go to the editing page, don't make any changes, just click "Cancel".
3. Go back to the question details page, click "Edit" under the same answer again.
4. Go to the editing page, and the main content disappears.

**Expected behavior**

In step 4, the answer's content should be displayed in the main text.

**Platform (please complete the following information):**

 - Browser Chrome
 - Version 1.2.1

hgaol:
fix #713 .

After changing to `useEffect`, the issue was resolved. I don't know why we use `useLayoutEffect`. Pls let me know if something I missed.

### DIFF
--- a/ui/src/pages/Questions/EditAnswer/index.tsx
+++ b/ui/src/pages/Questions/EditAnswer/index.tsx
@@ -73,7 +73,7 @@ const Index = () => {
   const [contentChanged, setContentChanged] = useState(false);
   const editCaptcha = useCaptchaModal('edit');
 
-  useLayoutEffect(() => {
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


shuashuai (approved):
It's true that `useLayoutEffect` is causing this bug, and it's related to the logic we used to refactor the editor's core functionality; thanks for contributing!

I will also check the logic of all used `useLayoutEffect` in V.1.2.5.

PR #717, opened 2024-01-09
Fix/1.2.5/unexpanded comments

insistedMarin:
**Title**: Improved Comment Sorting and Display Logic

**Description**:

This pull request introduces optimized sorting and display logic for comments. The focus is on enhancing user interaction with the comments section, particularly when dealing with varying numbers of comments.

**Key Updates**:

Top Comments Priority:

The top 3 highest-voted comments are displayed first, sorted by vote count and then by oldest posting time in case of a tie.
Handling Fewer Comments:

For scenarios with 3 or fewer comments, comments are sorted by posting time from oldest to newest after the database query. This approach was chosen due to the smaller data volume, enabling efficient post-query sorting.
Expandable Comments:

Added an "Expand Comments" feature. Post-expansion, comments are sorted by posting time from oldest to newest and loaded 15 at a time.
Dynamic Comment Count:

The expansion button shows the number of remaining comments to be viewed.
**Conclusion**:

These improvements aim to make the comment section more interactive and user-friendly, providing clarity and ease of access to the most relevant comments


internal/service/comment/comment_service.go
```diff
@@ -342,6 +342,11 @@ func (cs *CommentService) GetCommentWithPage(ctx context.Context, req *schema.Ge
 	if err != nil {
 		return nil, err
 	}
+	if total <= 3 {
+		sort.SliceStable(commentList, func(i, j int) bool {
+			return commentList[i].CreatedAt.Before(commentList[j].CreatedAt)
+		})
+	}
```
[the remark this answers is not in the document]
insistedMarin: Thank you very much for your guidance, I have solved the sorting problem on the front end.

ui/src/components/Comment/index.tsx
```diff
@@ -98,6 +98,9 @@ const Comment = ({ objectId, mode, commentId }) => {
       setComments([...comments, ...data.list]);
     }
     const user: Types.PageUser[] = [];
+    if (data.count <= 3) {
```
shuashuai, line 101: Should be sorted first and then assigned.

```
if (data.count <= 3) {
      data.list.sort((a, b) => a.created_at - b.created_at);
    }
  if (pageIndex === 1 || pageIndex === 0) {
      setComments(data?.list);
    } else {
      setComments([...comments, ...data.list]);
    }
    
    ... 
```

insistedMarin: Thank you very much, I have corrected the code.

ui/src/components/Comment/index.tsx
```diff
@@ -92,6 +92,9 @@ const Comment = ({ objectId, mode, commentId }) => {
     if (!data) {
       return;
     }
+    if (data.count <= 3) {
+      data.list.sort((a, b) => a.created_at - b.created_at);
+    }
     if (pageIndex === 1 || pageIndex === 0) {
```
shuashuai, line 98: Sorry, I missed some conditions yesterday.  `data.count` should be the total number of comments. Here you only need to sort the data loaded for the first time and with the same number of votes by time, so `data.list.length` and `vote_Count` should be used for judgment, and it should only be processed during the first load. So it should be like this

```javascript
  if (pageIndex === 1 || pageIndex === 0) {
      if (data.list.length <= 3) {
        data.list.sort((a, b) => {
          if (a.vote_count === b.vote_count) {
            return a.created_at - b.created_at;
          }
          return 1;
        });
      }
      setComments(data?.list);
    } else {
      setComments([...comments, ...data.list]);
    }
```

insistedMarin: image
Okay, there might be some confusion in the needs. I want to clarify the needs with you.

In my opinion, the main condition is to determine whether the total number of comments is greater than 3, and the comments less than or equal to 3 are displayed in order of time from old to new. For comments larger than 3, they will be displayed in two sortings initially and after expansion.

If the main condition is the number of comments obtained, then sorting from old to new should only be effective for cases less than 3. But the number of votes for this case is not a sorting condition.

In addition, the default sorting for "page 0" is votes from high to low and then time from old to new. So no need to reorder.

shuashuai: >  the main condition is to determine whether the total number of comments is greater than 3.

Your understanding is right.

>  In addition, the default sorting for "page 0" is votes from high to low and then time from old to new. So no need to reorder.

To be exact, the default sorting should not be based on "page 0", but you sort based on whether the total number is greater than 3, because "page 0" is a condition used by the front end to distinguish whether to load more,when using the useQueryComments` method, the parameter passed is page = 0, and the actual call passed is page=1.

The code is correct now！

Finally, thank you very much for your carefulness and contribution！

Issue #738, opened 2024-01-22
Bug Report: MeiliSearch Plugin Not Updating New Answers


hbsciw:
### Description
I am experiencing an issue where new answers are not updating in MeiliSearch when using the MeiliSearch plugin.

### Steps to Reproduce
1. Update the answers in the system.
2. Observe the behavior in MeiliSearch.

### Expected Behavior
New answers should be updated in MeiliSearch.

### Actual Behavior
New answers are not being updated in MeiliSearch.

### Debugging Information
While debugging, I encountered an error message stating "a pointer to a pointer is not allowed" at line 417 in `answer_repo.go`. Here is a screenshot of the error:

![Error screenshot](https://github.com/apache/incubator-answer/assets/7039555/ea9357e1-59b9-4352-ae11-b8af69884c03)

I suspect the issue might be related to the question instance initialization. As a workaround, I modified the initialization of the question instance, which seemed to resolve the issue temporarily. Here's the screenshot post-modification:

![Workaround screenshot](https://github.com/apache/incubator-answer/assets/7039555/6bcc3f4d-5b8a-4478-bb0e-32b71d4ff1bb)

### Question
Is this a known bug? Can anyone confirm if the issue is indeed related to the initialization of the question instance?

Any guidance or confirmation on this issue would be greatly appreciated.

kumfo:
@hbsciw Thanks for your feedback, you'r right. 

You can commit your change to fixed this bug.

// original code
```go
var (
    question *entity.Question
)
exist, err = ar.data.DB.Context(ctx).Where("id = ?", answer.QuestionID).Get(&question)
```

Can be modified according to your code.
```go
var (
    question = new(entity.Question)
)
exist, err = ar.data.DB.Context(ctx).Where("id = ?", answer.QuestionID).Get(&question)
```

or change like this:
``` go
var (
    question *entity.Question
)
exist, err = ar.data.DB.Context(ctx).Where("id = ?", answer.QuestionID).Get(question)
```

PR #754, opened 2024-01-27
fix: resolve MeiliSearch plugin not updating new answers

hbsciw:
Resolve issue with MeiliSearch plugin not updating new answers

Fixes bug reported in Issue #738

Description:
Addressed the issue where new answers were not updating in MeiliSearch when using the MeiliSearch plugin (also probabely for other search plugins). The problem was traced back to a pointer-related error in answer_repo.go (line 417). During debugging, an error message "a pointer to a pointer is not allowed" was encountered.

Debugging Information:
While investigating, I identified a potential problem with the initialization of the question instance. To mitigate the issue, I modified the initialization process.

##738 #bugfix

### DIFF
--- a/internal/repo/answer/answer_repo.go
+++ b/internal/repo/answer/answer_repo.go
@@ -416,7 +416,7 @@ func (ar *answerRepo) updateSearch(ctx context.Context, answerID string) (err er
 
 	// get question
 	var (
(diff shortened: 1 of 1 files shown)


kumfo (approved):
Good Job!

PR #776, opened 2024-02-06
fix: incomplete text display when pasting text into the tags field

### LINKED ISSUE #773: Incomplete text display when pasting text into the tags field
## Describe the bug

Paste the copied text to the tags field in ask question page, the word display is incomplete.

image

If you use the keyboard to type letter by letter, it can be displayed in its entirety.

image

### Expected behavior

Show full characters when pasted.

### Platform

- Device: Desktop
- OS: macOS
- Browser and version: Chrome
- Version: v1.2.5-RC2

hgaol:
fix #773 .

### DIFF
--- a/ui/src/components/TagSelector/index.tsx
+++ b/ui/src/components/TagSelector/index.tsx
@@ -320,7 +320,9 @@ const TagSelector: FC<IProps> = ({
         inputRef.current?.focus();
       }
     }
+  }, [focusState, tags, hiddenCreateBtn, searchValue, maxTagLength]);
 
+  useEffect(() => {
     // set width of tag Form.Control
     const ele = document.querySelector('.a-input-width') as HTMLElement;
     if (ele.offsetWidth > 60) {
@@ -331,7 +333,7 @@ const TagSelector: FC<IProps> = ({
     } else {
       inputRef.current?.setAttribute('style', 'width: 60px');
     }
-  }, [focusState, tags, hiddenCreateBtn, searchValue, maxTagLength]);
+  }, [searchValue]);
 
   return (
     <div ref={containerRef} className="position-relative">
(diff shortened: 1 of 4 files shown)


hgaol:
Hi @shuashuai , could you pls help to review this PR? Thanks a lot! You can only focus on `ui/src/components/TagSelector/index.tsx` because others are merged from `dev`.

ui/src/components/TagSelector/index.tsx
```diff
@@ -329,6 +320,17 @@ const TagSelector: FC<IProps> = ({
...
+    const ele = document.querySelector('.a-input-width') as HTMLElement;
+    if (ele.offsetWidth > 60) {
+      inputRef.current?.setAttribute(
+        'style',
+        `width:${ele.offsetWidth + 16}px`,
+      );
+    } else {
+      inputRef.current?.setAttribute('style', 'width: 60px');
+    }
   }, [focusState, tags, hiddenCreateBtn, searchValue, maxTagLength]);
```
shuashuai, line 334: The reason for the problem is that the actual width of the `.a-input-width` element cannot be obtained immediately after the value is assigned, so the best solution is to monitor the value of searchValue separately to avoid repeated calculations caused by changes in other values.

```
useEffect(() => {
    const ele = document.querySelector('.a-input-width') as HTMLElement;
    if (ele.offsetWidth > 60) {
      inputRef.current?.setAttribute(
        'style',
        `width:${ele.offsetWidth + 16}px`,
      );
    } else {
      inputRef.current?.setAttribute('style', 'width: 60px');
    }
  }, [searchValue]);

```

By the way, thanks for the contribution, when PRs are added to milestone, we'll work on those PRs before that version is released!

shuashuai (changes requested):

shuashuai (approved):

PR #800, opened 2024-02-23
fix: if PUBLIC_URL setted, static resources cannot be fetched from the backend.

liuyuanzhi-cn:
# Bug Reproduction Steps
1. Modify the `PUBLIC_URL` in the `ui/.env.production` file:
```text
  TSC_COMPILE_ON_ERROR=true
  ESLINT_NO_DEV_ERRORS=true
  PUBLIC_URL = /forum
  REACT_APP_API_URL = /
```
2. Recompile
```shell
  make ui
  make build
```
3. Execute the initialization command:
```shell
INSTALL_PORT=9080 ./answer init -C ./answer-data/
```
4. Access the initialization configuration page at `http://localhost:9080/install` and encounter the following error. 

answer error page

# Bug Resolution

During initialization and runtime, it is necessary to add the environment variable `PUBLIC_URL`.
```shell
# Initialization
INSTALL_PORT=9080 PUBLIC_URL=/forum ./answer init -C ./answer-data/
# Run
PUBLIC_URL=/forum ./answer run -C ./answer-data/
```

>Note：
>
>1. If the value of `PUBLIC_URL` is the default `/`, there is no need to set the environment variable.
>2. The value of `PUBLIC_URL` must not end with `/`. For example, `PUBLIC_URL=/forum` is correct, while `PUBLIC_URL=/forum/` is incorrect.

### DIFF
--- a/internal/base/constant/site_info.go
+++ b/internal/base/constant/site_info.go
@@ -48,5 +48,6 @@ const (
 )
 
 var (
-	PublicUrl = os.Getenv("PUBLIC_URL")
+	PublicUrl   = os.Getenv("PUBLIC_URL")
+	BaseUrlPath = os.Getenv("REACT_APP_BASE_URL_PATH")
 )
--- a/internal/base/constant/site_info.go
+++ b/internal/base/constant/site_info.go
@@ -19,6 +19,8 @@
 
 package constant
 
+import "os"
+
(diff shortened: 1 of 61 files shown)


shuashuai:
Related to this PR #836

PR #805, opened 2024-02-26
fix/slugify-non-ascii

hbsciw:
This pull request fixes an issue where questions with full Persian titles were causing a "many redirect error" when trying to access them through the link. The issue was caused by the slugify library returning a blank string for the title of these questions. With this fix, the slugify library has been updated to return a proper string for the title, resolving the issue and allowing for smooth access to all questions, regardless of their title.


hbsciw:
@LinkinStars Thank you for response. 
I solved 3 and 4.

1. A post with this title has many redirect problem.
`title= چگونه می‌توانم مساله گزارش شده را ارزبای کنم؟`
And post with this title hasn't problem.
`title = چگونه می‌توانم Bug گزارش شده را ارزیابی کنم؟ `

2. I think since the slugs aren't saved, it won't be a problem. I also tested a few things and it didn't seem to be a problem. Have you any specific scenario in your mind? Can you tell us your exact solution?

PR #822, opened 2024-03-04
fix: can't search username with '-' character

### LINKED ISSUE #819: Username Search Returning No Results
## Describe the bug
I try to search the user name in Answer, but failed.

### To Reproduce

1. Click Search.
2. Input: user:username.
3. No results shown.

### Expected behavior
Show posts, answers created by the user.

### Screenshots

https://github.com/apache/incubator-answer/assets/131739771/e2c21501-7c8d-4998-aa52-137c2099a353

### Platform

- Device: Desktop
- OS: macOS
- Browser and version: Chrome
- Version: 121.0.6167.85

hgaol:
Close #819 . Need @LinkinStars 's help to review if we could remove replacing '-' with ' ' for search, or any better solution.
image

### DIFF
--- a/internal/schema/search_schema.go
+++ b/internal/schema/search_schema.go
@@ -41,10 +41,32 @@ type SearchDTO struct {
 func (s *SearchDTO) Check() (errField []*validator.FormErrorField, err error) {
 	// Replace special characters.
 	// Special characters will cause the search abnormal, such as search for "#" will get nearly all the content that Markdown format.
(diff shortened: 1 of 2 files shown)


hgaol:
One possible solution is to just replace the search content instead of all the content. e.g. not replace pattern with key:val, [tag]. You can refer to the latest commit.

kumfo (approved):
Good job!

I have reviewed the subsequent underlying parsing logic, which includes rigorous secondary parsing and verification.  The key-value pairs here can allow these characters to pass through.

PR #830, opened 2024-03-06
fix(ui): fixed the ui issue of skeleton screen flickering

foxzero-007:
The changes made in this pull request are quite extensive, and the proposed solution is briefly described in the following text. I hope you can take some time to review it when you have a moment.

**What did I modify in this branch?**

Perhaps you have noticed that we have been using the skeleton screen technology on the frontend pages. However, the control over it seems a bit simplistic. We currently only initiate the skeleton screen when entering the page and directly replace it with new nodes upon interface returns. As a result, there is a "skeleton screen flickering" issue during usage. In favorable network conditions, we often observe a quick flash of the skeleton screen. From a user experience perspective, I prefer the page not to change so rapidly. Therefore, I have submitted this pull request to address the issue.

**How did I make the modification?**

Through code comparison, it can be observed that I actually encapsulated a new hook named `useSkeletonControl`. The parameters it returns are as follows:

```ts
{
     isSkeletonShow, // Should the skeleton screen be displayed?
     openSkeleton, // Activate the skeleton screen.
     closeSkeleton // Deactivate the skeleton screen.
}
```

In the code logic, you can see that I have implemented a delay-like operation for both activation and deactivation. The reason behind this is that I believe the opening logic of the skeleton screen should be like this:

- When the interface returns within 1 second, there is no need to activate the skeleton screen.
  
- When the interface takes more than 1 second to return, the skeleton screen is activated, and the activation time should not be less than 3 seconds.
  

(The above-mentioned times are all defined as constants and can be adjusted according to the actual situation.)

Based on this design approach, I utilized `setTimeout` and `clearTimeout` to control the timing of activation and deactivation.

Why provide a `needShowFirst` parameter?

Clearly, this parameter is used to control whether the skeleton screen is rendered right from the start. It allows the skeleton screen to bypass the first timing condition mentioned earlier, meaning it can be activated even if the interface returns in less than 1 second. The reason behind this is that I noticed in some pages, the skeleton screen needs to occupy its position right from the start to prevent the 'flickering' issue on the page. So, I introduced this parameter. However, in my design, I don't intend to show the skeleton screen when the interface response time is too short, as it might negatively impact the user experience.

The above is the reason behind my pull request. I have made modifications for the pages I identified, and I hope they will be accepted. Additionally, if there are better solutions, I am eager to see them. Thanks again for considering my changes.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -65,6 +65,7 @@ archives:
         dst: licenses/
       - src: "DISCLAIMER"
         dst: DISCLAIMER
+    wrap_in_directory: true
 checksum:
   name_template: 'checksums.txt'
 snapshot:
@@ -79,6 +80,7 @@ changelog:
 source:
   enabled: true
   name_template: apache-answer-{{ .RawVersion }}-incubating-src
+  prefix_template: "apache-answer-{{ .RawVersion }}-incubating-src/"
 
 # goreleaser release --skip-validate  --skip-publish --clean
 
--- a/i18n/cs_CZ.yaml
+++ b/i18n/cs_CZ.yaml
@@ -82,6 +82,9 @@ backend:
     level_3:
       description:
         other: Level 3 (high reputation required for mature community)
+    level_custom:
+      description:
+        other: Custom Level
     rank_question_add_label:
       other: Ask question
     rank_answer_add_label:
@@ -681,6 +684,12 @@ ui:
             character: URL slug contains unallowed character set.
         desc:
           label: Description
+        revision:
+          label: Revision
+        edit_summary:
+          label: Edit summary
+          placeholder: >-
+            Briefly explain your changes (corrected spelling, fixed grammar, improved formatting)
     btn_cancel: Cancel
     btn_submit: Submit
     btn_post: Post new tag
@@ -708,21 +717,6 @@ ui:
     title: Edit Tag
     default_reason: Edit tag
     default_first_reason: Add tag
-    form:
-      fields:
-        revision:
-          label: Revision
-        display_name:
-          label: Display name
-        slug_name:
-          label: URL slug
-          info: URL slug up to 35 characters.
-        desc:
-          label: Description
-        edit_summary:
-          label: Edit summary
-          placeholder: >-
-            Briefly explain your changes (corrected spelling, fixed grammar, improved formatting)
     btn_save_edits: Save edits
     btn_cancel: Cancel
   dates:
@@ -746,7 +740,7 @@ ui:
     btn_flag: Flag
     btn_save_edits: Save edits
     btn_cancel: Cancel
-    show_more: Show more comments
+    show_more: "{{count}} more comments"
     tip_question: >-
       Use comments to ask for more information or suggest improvements. Avoid answering questions in comments.
     tip_answer: >-
@@ -971,10 +965,10 @@ ui:
         label: Location
         placeholder: "City, Country"
     notification:
-      heading: Notifications
-      email: Email
+      heading: Email Notifications
+      turn_on: Turn on
       inbox:
-        label: Email notifications
+        label: Inbox notifications
         description: Answers to your questions, comments, invites, and more.
       all_new_question:
         label: All new questions
@@ -1029,7 +1023,7 @@ ui:
     answers: answers
   invite_to_answer:
     title: People Asked
-    desc: Invite people who you think might know the answer.
+    desc: Select people who you think might know the answer.
     invite: Invite to answer
     add: Add people
     search: Search people
@@ -1123,6 +1117,11 @@ ui:
     close: Close
     reopen: Reopen
     ok: OK
+    light: Light
+    dark: Dark
+    system_setting: System setting
+    default: Default
+    reset: Reset
   search:
     title: Search Results
     keywords: Keywords
@@ -1642,11 +1641,12 @@ ui:
       themes:
         label: Themes
         text: Select an existing theme.
+      color_scheme:
+        label: Color scheme
       navbar_style:
-        label: Navbar Style
-        text: Select an existing theme.
+        label: Navbar style
       primary_color:
-        label: Primary Color
+        label: Primary color
         text: Modify the colors used by your themes
     css_and_html:
       page_title: CSS and HTML
--- a/i18n/cy_GB.yaml
+++ b/i18n/cy_GB.yaml
@@ -82,6 +82,9 @@ backend:
     level_3:
       description:
         other: Level 3 (high reputation required for mature community)
+    level_custom:
+      description:
+        other: Custom Level
     rank_question_add_label:
       other: Ask question
     rank_answer_add_label:
@@ -681,6 +684,12 @@ ui:
             character: Mae slug URL yn cynnwys set nodau na caniateir.
         desc:
           label: Disgrifiad
+        revision:
+          label: Revision
+        edit_summary:
+          label: Edit summary
+          placeholder: >-
+            Briefly explain your changes (corrected spelling, fixed grammar, improved formatting)
     btn_cancel: Canslo
     btn_submit: Cyflwyno
     btn_post: Post tag newydd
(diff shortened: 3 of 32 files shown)


shuashuai:
Thank you very much for your contribution. We will first discuss the advantages, disadvantages and necessity of this solution. Please be patient.

foxzero-007:
> Thank you very much for your contribution. We will first discuss the advantages, disadvantages and necessity of this solution. Please be patient.

Ok, thank you for your reply. Look forward to hearing from you.

robinv8:
I understand your thinking, but I think it complicates the problem. I have a suggestion:

Create a Wrapper component, and then use setTimeout to control whether to display the skeleton screen, for example
```
<ListWrapper isLoading={isLoading} loader={<QuestionListLoader/>}>
....
</ListWrapper>
```

foxzero-007:
> I understand your thinking, but I think it complicates the problem. I have a suggestion:
> 
> Create a Wrapper component, and then use setTimeout to control whether to display the skeleton screen, for example
> 
> ```
> <ListWrapper isLoading={isLoading} loader={<QuestionListLoader/>}>
> ....
> </ListWrapper>
> ```

Thank you very much for your response. Here, I would like to share my thoughts.

In fact, whether it's `components` or `hooks`, our core lies in handling the `loading state`. From my understanding, there isn't a particularly significant difference in the implementation of `loading` between the two approaches. Personally, in functional components, I prefer using hooks to manage data exclusively and abstracting the management of the `loading state`'out of the component. In contrast, if we use a `component-based` approach, we need to encapsulate a portion of the `rendering logic`. I am uncertain whether the rendering logic in the `ListWrapper` component will be generic enough to meet all business scenarios in future situations.

robinv8:
> > I understand your thinking, but I think it complicates the problem. I have a suggestion:
> > Create a Wrapper component, and then use setTimeout to control whether to display the skeleton screen, for example
> > ```
> > <ListWrapper isLoading={isLoading} loader={<QuestionListLoader/>}>
> > ....
> > </ListWrapper>
> > ```
> 
> Thank you very much for your response. Here, I would like to share my thoughts.
> 
> In fact, whether it's `components` or `hooks`, our core lies in handling the `loading state`. From my understanding, there isn't a particularly significant difference in the implementation of `loading` between the two approaches. Personally, in functional components, I prefer using hooks to manage data exclusively and abstracting the management of the `loading state` out of the component. In contrast, if we use a `component-based` approach, we need to encapsulate a portion of the `rendering logic`. I am uncertain whether the rendering logic in the `ListWrapper` component will be generic enough to meet all business scenarios in future situations.

I agree with using hooks, but I don’t recommend exposing too much logic to business components. Secondly, the implementation of `useSkeletonControl` is too complicated, and the settings of `SKELETON_SHOW_MIN_TIME` and `SKELETON_NEED_TIME` are also unfounded. Under normal circumstances, the interface The response speed is at the millisecond level and does not need to be so complicated.

foxzero-007:
> > > I understand your thinking, but I think it complicates the problem. I have a suggestion:
> > > Create a Wrapper component, and then use setTimeout to control whether to display the skeleton screen, for example
> > > ```
> > > <ListWrapper isLoading={isLoading} loader={<QuestionListLoader/>}>
> > > ....
> > > </ListWrapper>
> > > ```
> > 
> > 
> > Thank you very much for your response. Here, I would like to share my thoughts.
> > In fact, whether it's `components` or `hooks`, our core lies in handling the `loading state`. From my understanding, there isn't a particularly significant difference in the implementation of `loading` between the two approaches. Personally, in functional components, I prefer using hooks to manage data exclusively and abstracting the management of the `loading state` out of the component. In contrast, if we use a `component-based` approach, we need to encapsulate a portion of the `rendering logic`. I am uncertain whether the rendering logic in the `ListWrapper` component will be generic enough to meet all business scenarios in future situations.
> 
> I agree with using hooks, but I don’t recommend exposing too much logic to business components. Secondly, the implementation of `useSkeletonControl` is too complicated, and the settings of `SKELETON_SHOW_MIN_TIME` and `SKELETON_NEED_TIME` are also unfounded. Under normal circumstances, the interface The response speed is at the millisecond level and does not need to be so complicated.



> > > I understand your thinking, but I think it complicates the problem. I have a suggestion:
> > > Create a Wrapper component, and then use setTimeout to control whether to display the skeleton screen, for example
> > > ```
> > > <ListWrapper isLoading={isLoading} loader={<QuestionListLoader/>}>
> > > ....
> > > </ListWrapper>
> > > ```
> > 
> > 
> > Thank you very much for your response. Here, I would like to share my thoughts.
> > In fact, whether it's `components` or `hooks`, our core lies in handling the `loading state`. From my understanding, there isn't a particularly significant difference in the implementation of `loading` between the two approaches. Personally, in functional components, I prefer using hooks to manage data exclusively and abstracting the management of the `loading state` out of the component. In contrast, if we use a `component-based` approach, we need to encapsulate a portion of the `rendering logic`. I am uncertain whether the rendering logic in the `ListWrapper` component will be generic enough to meet all business scenarios in future situations.
> 
> I agree with using hooks, but I don’t recommend exposing too much logic to business components. Secondly, the implementation of `useSkeletonControl` is too complicated, and the settings of `SKELETON_SHOW_MIN_TIME` and `SKELETON_NEED_TIME` are also unfounded. Under normal circumstances, the interface The response speed is at the millisecond level and does not need to be so complicated.

Thanks for your reply,
I strongly agree that a skeleton screen should not be displayed when the interface returns very quickly. However, currently, we are facing issues with the control strategy for the skeleton screen when the interface response is too fast. This is why I need to implement a more complex solution to control the loading. The meanings of the two parameters mentioned above are as follows:

- `SKELETON_NEED_TIME`:Define when the skeleton screen should be displayed after the interface response exceeds a certain duration.
- `SKELETON_SHOW_MIN_TIME`:The minimum duration for displaying the skeleton screen (I believe the skeleton screen should be maintained for a certain period once displayed to prevent it from flashing too quickly).

When the interface response time is less than `SKELETON_NEED_TIME`, we can directly hide the skeleton screen without displaying it. However, when the response time is greater than `SKELETON_NEED_TIME`, I prefer the skeleton screen to be displayed and maintained for a certain duration. Regarding the current assignments for these two parameters, I've set them arbitrarily. In reality, these values should be determined by UI designers as they represent the user's psychological expectations for response times.

In fact, this logic only delays the switching time of the loading state to achieve control over the display of the skeleton screen. It should not involve any business logic but rather exists purely as a tool. I'm not quite sure about the use of `complicated` in this context as it seems to be a straightforward implementation without intricate business logic.

robinv8:
@foxzero-007 Hello, I have added suggestions for rectification and look forward to your reply.

foxzero-007:
> @foxzero-007 Hello, I have added suggestions for rectification and look forward to your reply.

Thanks for your reply,
I strongly agree that a skeleton screen should not be displayed when the interface returns very quickly. However, currently, we are facing issues with the control strategy for the skeleton screen when the interface response is too fast. This is why I need to implement a more complex solution to control the loading. The meanings of the two parameters mentioned above are as follows:

SKELETON_NEED_TIME:Define when the skeleton screen should be displayed after the interface response exceeds a certain duration.
SKELETON_SHOW_MIN_TIME:The minimum duration for displaying the skeleton screen (I believe the skeleton screen should be maintained for a certain period once displayed to prevent it from flashing too quickly).
When the interface response time is less than SKELETON_NEED_TIME, we can directly hide the skeleton screen without displaying it. However, when the response time is greater than SKELETON_NEED_TIME, I prefer the skeleton screen to be displayed and maintained for a certain duration. Regarding the current assignments for these two parameters, I've set them arbitrarily. In reality, these values should be determined by UI designers as they represent the user's psychological expectations for response times.

In fact, this logic only delays the switching time of the loading state to achieve control over the display of the skeleton screen. It should not involve any business logic but rather exists purely as a tool. I'm not quite sure about the use of complicated in this context as it seems to be a straightforward implementation without intricate business logic.

robinv8:
Hello, thank you very much for your contribution. After discussion with our team, we still need you to make rectifications according to the results of the review, otherwise it will not be merged.

foxzero-007:
@robinv8 Okay, I have simplified the implementation of the hook. Perhaps you can take a look and see if this solution is feasible.

robinv8:
Hello, I have simplified it based on your code, see here
https://github.com/foxzero-007/incubator-answer/pull/1

PR #836, opened 2024-03-11
Feat/base url config

liuyuanzhi-cn:
# Problem
I would love to use the second-level domain name of nginx to deploy `Answer`. However, this requires the  webapp to work on relative url path (e.g. /foo/questions) or support setting a base url via configuration.

# Solution
Add support for setting a base/root url in the configuration or make the webapp work on relativ urls. Having this option is very useful in a variety of deployment scenarios (e.g. behind a reverse proxy).

Adds the config option `REACT_APP_BASE_URL_PATH` . When set to a value like `/foo`, Anwser will be served from `hostname:port/foo`.

# Usage

1. edit the `ui/.env.production` file
```text
...
REACT_APP_BASE_URL_PATH=/foo
...
```
2. Recompile
```shell
make ui
make build
```

3. Initialization and Run
```shell
# Initialization
INSTALL_PORT=9080 REACT_APP_BASE_URL_PATH=/foo ./answer init -C ./answer-data/
# Run
REACT_APP_BASE_URL_PATH=/foo ./answer run -C ./answer-data/
```

>Note：
>
>1. If the value of `REACT_APP_BASE_URL_PATH` is the default `""`, there is no need to set the environment variable.
>2. The value of `REACT_APP_BASE_URL_PATH` must not end with `/`. For example, `PUBLIC_URL=/foo` is correct, while `PUBLIC_URL=/foo/` is incorrect.

### DIFF
--- a/internal/base/constant/site_info.go
+++ b/internal/base/constant/site_info.go
@@ -48,5 +48,6 @@ const (
 )
 
 var (
-	PublicUrl = os.Getenv("PUBLIC_URL")
+	PublicUrl   = os.Getenv("PUBLIC_URL")
+	BaseUrlPath = os.Getenv("REACT_APP_BASE_URL_PATH")
 )
--- a/internal/base/constant/site_info.go
+++ b/internal/base/constant/site_info.go
@@ -19,6 +19,8 @@
 
 package constant
 
+import "os"
+
 const (
 	DefaultGravatarBaseURL = "https://www.gravatar.com/avatar/"
 	DefaultAvatar          = "system"
@@ -43,4 +45,8 @@ const (
 	ColorSchemeLight   = "light"
 	ColorSchemeDark    = "dark"
 	ColorSchemeSystem  = "system"
+)
+
+var (
+	PublicUrl = os.Getenv("PUBLIC_URL")
 )
--- a/internal/install/install_controller.go
+++ b/internal/install/install_controller.go
@@ -20,6 +20,7 @@
 package install
 
 import (
+	"github.com/apache/incubator-answer/internal/base/constant"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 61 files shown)


shuashuai:
react-router-dom supports adding basename, which will save a lot of files. by adjusting like this, there should be less code modifications.

```
 const router = createBrowserRouter(routes, {
    basename: process.env.REACT_APP_BASE_URL_PATH,
  });
```

However, by adding basename, no matter how you adjust it, there will be a certain degree of code intrusion, and a lot of testing is required, because it may also include some callback addresses returned by the API, or the oauth address in the plug-in, etc.

Another way is to directly use nginx as a proxy without changing the code in the business. How effective is this method?

shuashuai:
@liuyuanzhi-cn  

I looked at the changes you submitted. There will be a lot of code intrusion and modification. I tried to achieve it by adding basename in combination with react-router-dom. This can be achieved with only a small amount of changes. You can refer to [here](https://github.com/apache/incubator-answer/tree/feat/baseurl).

Of course, I only modified the front-end related routing configuration, and some routes returned through the API have not yet been adjusted. This modification should be reasonable.

liuyuanzhi-cn:
> @liuyuanzhi-cn
> 
> I looked at the changes you submitted. There will be a lot of code intrusion and modification. I tried to achieve it by adding basename in combination with react-router-dom. This can be achieved with only a small amount of changes. You can refer to [here](https://github.com/apache/incubator-answer/tree/feat/baseurl).
> 
> Of course, I only modified the front-end related routing configuration, and some routes returned through the API have not yet been adjusted. This modification should be reasonable.

Indeed, this looks great, it's much more elegant this way, with a lot less code changes.

shuashuai:
> > @liuyuanzhi-cn
> > I looked at the changes you submitted. There will be a lot of code intrusion and modification. I tried to achieve it by adding basename in combination with react-router-dom. This can be achieved with only a small amount of changes. You can refer to [here](https://github.com/apache/incubator-answer/tree/feat/baseurl).
> > Of course, I only modified the front-end related routing configuration, and some routes returned through the API have not yet been adjusted. This modification should be reasonable.
> 
> Indeed, this looks great, it's much more elegant this way, with a lot less code changes.

We will confirm support for this requirement in subsequent versions as soon as possible! Of course, if you are interested, you are welcome to continue to submit PR according to this idea.

shuashuai:
We will support this feature in version 1.3.5. The PR you contributed here is currently not applicable, so I will close the PR first.

PR #882, opened 2024-03-26
fix: Account setting label issue

### LINKED ISSUE #881: Account setting label issue
## Describe the bug

image

This is the current email, not the new email.

### To Reproduce

Visit /users/settings/account

### Expected behavior

Change label text from "New email" to "Email".

### Platform

- Version: v1.2.5

sy-records:
fix #881

### DIFF
--- a/i18n/af_ZA.yaml
+++ b/i18n/af_ZA.yaml
@@ -715,7 +715,9 @@ ui:
         We've sent an email to that address. Please follow the confirmation instructions.
       email:
         label: Email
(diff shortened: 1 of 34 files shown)


shuashuai (approved):
Thanks for the contribution. If you don’t want to translate other languages, you can just modify the `en_US.yaml` file, because we use [crowdin](https://crowdin.com/project/answer) to manage translation. Other language files are generated based on en_US. Changes in this before reversion will be automatically synchronized to other language files!

Issue #887, opened 2024-04-01
No error message is showing in UI - when user update comment and comment exceeds comment max length


surapuramakhil:
## Describe the bug

Comment update doesn't save when comment length exceeds its limits. No error message is showing in UI

On Inspect:
400 Bad Request is returned when comment size exceeds its max length.

## Suggestion
To show error message - maximum comment size in error pop.

### Screenshots

If applicable, add screenshots or video to help explain your problem.

### Platform

- Device: [e.g. Desktop, Mobile]
- OS: [e.g. macOS]
- Browser and version: [e.g. Chrome, Safari]
- Version: [e.g. v1.2.0]

surapuramakhil:
Ah, this only comment on new comment but doesn't come when users update their comments.

I will update this.

zahash:
@LinkinStars made a pr. works as expected. https://github.com/apache/incubator-answer/pull/893

PR #893, opened 2024-04-05
fix bug where no error message is showing when comment length exceeds max length when updating comment

zahash:
fixes https://github.com/apache/incubator-answer/issues/887


shuashuai (approved):
Thanks for the contribution

PR #918, opened 2024-04-23
Fix 909 Update initial data

zahash:


zahash:
@LinkinStars 

one of the tests fails because it has a `tag_rel` with hardcoded id that clashed with one present in initial data. 
i will fix the test soon by changin the hardcoded it to something else.

zahash:
@LinkinStars 

i think i resolved all the code change requests. can you check?

Issue #916, opened 2024-04-23
The button miss text in the mobile view


fenbox:
## Describe the bug

The button miss text in Admin->Content->Questions in the mobile view.

### Expected behavior

text: More

### Screenshots

![image](https://github.com/apache/incubator-answer/assets/360618/722e50d9-ebec-4b3e-8f91-2da43228dd34)

### Platform

- Device: Mobile
- OS: iOS
- Browser and version: Safari
- Version: v1.3.0

zahash:
@fenbox 

i'll do this

zahash:
@fenbox 

here is the PR https://github.com/apache/incubator-answer/pull/921

PR #921, opened 2024-04-24
Fix 916

zahash:
fixes https://github.com/apache/incubator-answer/issues/916


shuashuai:
Thank you for your contribution!  @sy-records #922  method is correct!

PR #953, opened 2024-05-20
feat: Support SMTP Encryption using TLS

### LINKED ISSUE #879: SMTP not support outlook
## Describe the bug

SMTP have  no option with TLS, so I can't use Outlook as my email sender.

### To Reproduce

1. logging as admin
2. Go to setting-> SMTP -> Encryption 
3. use SSL or None (no TLS option)
4. fill in outlook  smtp server
5. click  save

### Expected behavior
no test email can receive.
but when i use qq email as my email sender, it's work normal.
so i'm sure, outlook not work because not support TSL.
also outlook clear demand TLS.

### Platform

- Device: [Docker]
- OS: [Unraid OS]
- Browser and version: [Chrome, 122.0.6261.129]
- Version: [v1.2.5]

sy-records:
close #879

### DIFF
--- a/docs/docs.go
+++ b/docs/docs.go
@@ -6208,41 +6208,6 @@ const docTemplate = `{
                 }
             }
         },
-        "/answer/api/v1/user/register/captcha": {
-            "get": {
-                "description": "UserRegisterCaptcha",
-                "consumes": [
-                    "application/json"
-                ],
-                "produces": [
-                    "application/json"
-                ],
-                "tags": [
-                    "User"
-                ],
-                "summary": "UserRegisterCaptcha",
-                "responses": {
-                    "200": {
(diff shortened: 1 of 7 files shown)


sy-records:
good catch. I tested it and sent it properly, so I missed it here.

internal/service/export/email_service.go
```diff
@@ -135,6 +139,9 @@ func (es *EmailService) Send(ctx context.Context, toEmailAddr, subject, body str
 	if ec.IsSSL() {
 		d.SSL = true
 	}
+	if ec.IsTLS() {
+		d.SSL = false
```
sy-records, line 143: see https://github.com/go-gomail/gomail/issues/135#issuecomment-986031065

PR #955, opened 2024-05-27
ko_KR.yaml update

kimhsno1:
Add new translations.
English to Korean

### DIFF
--- a/i18n/ko_KR.yaml
+++ b/i18n/ko_KR.yaml
@@ -16,7 +16,9 @@
 # under the License.
 
 # The following fields are used for back-end
+
 backend:
+
   base:
     success:
       other: 성공.
@@ -34,7 +36,7 @@ backend:
       other: 중복된 제출입니다.
   action:
     report:
-      other: Flag
+      other: 신고하기
     edit:
       other: 편집
     delete:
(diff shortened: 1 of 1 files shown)


i18n/ko_KR.yaml
```diff
@@ -19,1851 +19,1846 @@
...
-      text: Chart
-      flow_chart: Flow chart
-      sequence_diagram: Sequence diagram
-      class_diagram: Class diagram
-      state_diagram: State diagram
-      entity_relationship_diagram: Entity relationship diagram
-      user_defined_diagram: User defined diagram
-      gantt_chart: Gantt chart
-      pie_chart: Pie chart
+    pagination:
```
shuashuai, line 505: There is a problem with the indentation of the following paragraph

shuashuai (changes requested):
First of all, thank you very much for your contribution. The yaml file has strict requirements for indentation, so I will not check them one by one. Please compare the `en_US.yaml`  file carefully and adjust the corresponding indentation.I just pointed out a few issues.

For example, this kind of large-scale adjustment is basically a problem of indentation.
image

If you are not a technical person, you can also go to [crowdin](https://crowdin.com/project/answer) to translate the file. You don’t need to pay attention to the file format here.

PR #965, opened 2024-05-31
fix: loginToContinueModal z-index error

### LINKED ISSUE #963: Click "Follow" on the question page, and the prompt text of the "Follow" button will be displayed on the login box.
## Describe the bug

Click "Follow" on the question page, and the prompt text of the "Follow" button will be displayed on the login box.

image

### Platform

- Device: iPhone 13 (15.1)

### LINKED ISSUE #964: The modal area don't need a title.
## Describe the bug

The modal area `<div>` don't need a title.

### Expected behavior

Remove `title="..."`

### Screenshots

image

### Platform

- Version: v1.3.1

sy-records:
fix #963 
fix #964

### DIFF
--- a/ui/src/pages/Questions/Detail/components/Reactions/index.tsx
+++ b/ui/src/pages/Questions/Detail/components/Reactions/index.tsx
@@ -50,6 +50,7 @@ const Index: FC<Props> = ({
 
   const handleSubmit = (params: { object_id: string; emoji: string }) => {
     if (!tryNormalLogged(true)) {
+      setReactIsActive(false);
       return;
     }
     updateReaction({
@@ -135,7 +136,7 @@ const Index: FC<Props> = ({
               }>
               <Button
                 title={emoji.name}
-                className="rounded-pill ms-2 link-secondary d-flex align-items-center"
+                className="rounded-pill ms-2 link-secondary align-items-center"
                 variant="light"
                 size="sm"
                 onClick={() =>
--- a/ui/src/components/Comment/index.tsx
+++ b/ui/src/components/Comment/index.tsx
@@ -364,9 +364,12 @@ const Comment = ({ objectId, mode, commentId }) => {
   };
 
   const handleAddComment = () => {
-    if (tryNormalLogged(true)) {
-      setVisibleComment(!visibleComment);
+    if (!tryNormalLogged(true)) {
+      setVisibleComment(false);
+      return;
     }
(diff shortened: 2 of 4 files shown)


ui/src/pages/Questions/Detail/components/Reactions/index.tsx
```diff
@@ -135,7 +136,7 @@ const Index: FC<Props> = ({
               }>
               <Button
                 title={emoji.name}
-                className="rounded-pill ms-2 link-secondary d-flex align-items-center"
+                className="rounded-pill ms-2 link-secondary align-items-center"
```
sy-records, line 139: Fix style on mobile.

| before | after |
| --- | --- |
|  image |  image |

ui/src/common/loginToContinueModal.scss
```diff
@@ -0,0 +1,24 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+@media screen and (max-width:575.98px) {
+  div[role='dialog'] {
+    z-index: 1100;
+  }
+}
```
shuashuai, line 24: In fact, this should be regarded as a global bug, because the z-index attribute of tootip defined in bootstrap is higher than that of modal, so I think we should increase the z-index of modal or lower the z-index of tooltip.

bootstrap defined CSS

image

You can define variables in the `src/common/_variable.scss` directory

image

shuashuai (changes requested):

shuashuai:
Sorry I didn't think it all through last week, and the problem has been on bootstrap for a long time without a good solution, see [here](https://github.com/twbs/bootstrap/issues/35202)

We discussed this issue yesterday, and realized that globally overriding bootstrap variables causes a problem when the tooltip is used in a modal, the tooltip will be obscured. So in order to reduce the scope of influence, we decided to only increase the modal's z-index when the login dialog box appears.

It's more like your first program.
1. create `login.scss` under the ui/src/components/Modal folder, and import in `LoignToContinueModal.tsxt` file
2. `login.scss` content
```
.modal-backdrop  {
  z-index: 1081;
}

.loginToContinueModal.show {
  z-index: 1082;
}

```
`LoignToContinueModal.tsxt` add className attribute

image

shuashuai (approved):
Thank you for your contribution！

Issue #961, opened 2024-05-30
404 error occurs, when use Chinese characters as tag


byerer:
## Describe the bug
When attempting to access the URL "https://www.example.com/tags/..." with "中文" (Chinese) used as a tag, a 404 error occurs.

### To Reproduce
![1](https://github.com/apache/incubator-answer/assets/110678099/146bb15c-434b-4e61-85c2-92d805dad33b)
1. If I directly click on the Chinese tag, it will jump to url like this.
![2](https://github.com/apache/incubator-answer/assets/110678099/d29f1b19-e6da-4b34-af13-30da0135ef5f)
3. So 404 happened.

### Expected behavior

200 response

### Source
[404 error occurs, when use Chinese characters as tag](https://meta.answer.dev/questions/D1OO2/404-error-occurs-when-use-chinese-characters-as-tag)

byerer:
I'd be happy to help with this issue. :)

PR #970, opened 2024-06-03
fix: Chinese character tag bug

byerer:
#961 

Adopted the method used for handling UrlTitle to process slugname.


byerer:
Initially, I also thought it was a front-end issue, because the encodeURIComponent function was used, so I deleted `encodeURICompontent` in the functions related to the tag `slugName`.

For example, in   `ui/src/router/pathFactory.ts`
```
const tagLanding = (slugName: string) => {
  const r = slugName ? `/tags/${encodeURIComponent(slugName)}` : '/tags';
  return r;
};
```

Also in `ui/src/components/Tag/index.tsx`
```
const Index: FC<IProps> = ({
  ...
  href ||= pathFactory.tagLanding(encodeURIComponent(data.slug_name));
  ...
})
```

This will lead to new issues; if the slugName contains '/' or some special characters, it will result in URI errors.

Thus, I decided to resolve this issue using a method similar to the one used for handling `UrlTitle`.

I'm not sure if my understanding is correct, and I would be happy to discuss this issue with you.

shuashuai:
> Initially, I also thought it was a front-end issue, because the encodeURIComponent function was used, so I deleted `encodeURICompontent` in the functions related to the tag `slugName`.
> 
> For example, in `ui/src/router/pathFactory.ts`
> 
> ```
> const tagLanding = (slugName: string) => {
>   const r = slugName ? `/tags/${encodeURIComponent(slugName)}` : '/tags';
>   return r;
> };
> ```
> 
> Also in `ui/src/components/Tag/index.tsx`
> 
> ```
> const Index: FC<IProps> = ({
>   ...
>   href ||= pathFactory.tagLanding(encodeURIComponent(data.slug_name));
>   ...
> })
> ```
> 
> This will lead to new issues; if the slugName contains '/' or some special characters, it will result in URI errors.
> 
> Thus, I decided to resolve this issue using a method similar to the one used for handling `UrlTitle`.
> 
> I'm not sure if my understanding is correct, and I would be happy to discuss this issue with you.

Tags are different from titles. tags are created without any checksums, so in order to prevent special characters from damaging the integrity of the URL, `encodeURIComponent` is used here to escape.

this was done with the following main considerations in mind:

1. make the display of the tag on the page look the same as when it was created.
2. no need to consider the compatibility of some of its old data.

like this:
image
image

The reason for using the `encodeURIComponent` method is to ensure that the URL conforms to the requirements, and there is no such restriction in the API, so we just need to restore the escaped parameters in the API inputs.

The APIs involved are

1. `/answer/api/v1/tag?name=xxxx`
2. `/answer/api/v1/question/page?page_size=20&page=1&order=active&tag=xxxx`

If you have other questions, you can continue the discussion.

PR #971, opened 2024-06-03
fix: Chinese character tag bug

byerer:
#961 

The reason for the error handling of escaped characters in the `tag component` is due to the redundant use of the encodeURIComponent function when passing the slug parameter. The `slugName` has already been processed in `pathFactory.tagLanding`, so it does not need to be processed again when passing parameters.

I'm sorry for the incorrect operations on my Git.

Thanks for your help @LinkinStars @shuashuai


shuashuai (approved):
Indeed. Thank you for your careful scrutiny.

PR #1028, opened 2024-07-19
feat: Support admin edit display name

### LINKED ISSUE #1024: Admin able to update DisplayName of user
## Is your feature request related to a problem? Please describe

We use SSO to log in, if the user is renamed in AD, the following changes must be made (otherwise a new user will be created):
email, username, displayname

It would be nice to change DisplayName from the admin interface.

Changing the display name, user name and email should also be possible via API.

I (and the user) are currently making the changes manually and before the user logs in again, which is very error prone.

sy-records:
fix #1024

### DIFF
--- a/docs/docs.go
+++ b/docs/docs.go
@@ -2985,7 +2985,7 @@ const docTemplate = `{
                                     "type": "object",
                                     "properties": {
                                         "data": {
-                                            "$ref": "#/definitions/schema.ReactionResp"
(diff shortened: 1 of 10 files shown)


shuashuai:
If you edit the same user repeatedly, the user information will display old data.

image

Edit again
image

shuashuai (approved):
Thanks for the contribution. The front-end code has been reviewed and is waiting for the back-end to be reviewed.

Issue #1093, opened 2024-09-03
Error 1055 in ListNewestEarned


lhpqaq:
### To Reproduce

1. Go to Homepage
2. Click on Profile in the upper right corner
// 3. Scroll down to '....'
```
2024-09-03 11:18:06.145 ERROR   handler/handler.go:45   Error 1055: Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'answer.badge_award.user_id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
/Users/lhp/Project/incubator-answer/internal/controller/badge_controller.go:158 github.com/apache/incubator-answer/internal/controller.(*BadgeController).GetRecentBadgeAwardListByUsername
/Users/lhp/go/pkg/mod/github.com/gin-gonic/gin@v1.9.1/context.go:174 github.com/gin-gonic/gin.(*Context).Next
/Users/lhp/Project/incubator-answer/internal/base/middleware/auth.go:88 github.com/apache/incubator-answer/internal/base/server.NewHTTPServer.(*AuthUserMiddleware).EjectUserBySiteInfo.func9
```
### Screenshots

<img width="1038" alt="image" src="https://github.com/user-attachments/assets/558ce7bf-d74f-4e25-b842-aaf4b939a917">

### Platform

- Device: [e.g. Desktop, Mobile]
- OS: macOS
- Browser and version: chrome
- Version: go version go1.22.2 darwin/arm64, mysql  Ver 9.0.0 for macos14 on arm64

The reason may be that in higher versions of MySQL, when executing group by, if the select field does not belong to the group by field, the SQL statement will report error 1055.

PR #1094, opened 2024-09-03
fix Error 1055 in ListNewestEarned

lhpqaq:
In higher versions of MySQL, when executing group by, if the select field does not belong to the group by field, the SQL statement will report error 1055. 
See #1093


kumfo (approved):
Good job! 
Thanks.

Issue #1077, opened 2024-08-21
Installation page supports i18n


sosyz:
## Is your enhancement request related to a problem? Please describe

When we first run answer we need to configure the database connection information on the installation page, but this page only supports English and does not change the UI language with the language option.

<img width="1107" alt="image" src="https://github.com/user-attachments/assets/03d03d48-e8a9-4a2a-b53d-44ec8af82dbe">
<img width="801" alt="image" src="https://github.com/user-attachments/assets/49b870f3-530d-40fd-80f4-c39e91a22b68">

## Describe the solution you'd like

Change the interface display language when changing language options.

PR #1105, opened 2024-09-11
feat: install page support i18n

sy-records:
fix https://github.com/apache/incubator-answer/issues/1077#issuecomment-2342970839

### DIFF
--- a/ui/src/pages/Install/index.tsx
+++ b/ui/src/pages/Install/index.tsx
@@ -280,6 +280,7 @@ const Index: FC = () => {
           db_connection_success: res.db_connection_success,
         });
         if (res && res.config_file_exist) {
+          setupInstallLanguage(Storage.get(CURRENT_LANG_STORAGE_KEY));
           if (res.db_connection_success) {
             setStep(6);
           } else {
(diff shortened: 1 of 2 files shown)


ui/src/pages/Install/index.tsx
```diff
@@ -294,7 +295,6 @@ const Index: FC = () => {
 
   useEffect(() => {
     configYmlCheck();
-    setupInstallLanguage(Storage.get(CURRENT_LANG_STORAGE_KEY));
   }, []);
```
shuashuai, line 298: No need to adjust here.

You only need to adjust the FirstStep component and determine whether there is local storage for the language when the component is initialized, that is,

eg:
image

Because the first value is taken by default here, it will not match the cached options.

shuashuai (changes requested):

ui/src/pages/Install/index.tsx
sy-records: Using currentLang does not switch to that language by default.

This is just to show the correct language when the config file exists and step 6 or 7 is executed. So moved to configYmlCheck.

shuashuai: This processing method does not meet the expectations of #1077, Maybe you can discuss with @LinkinStars 
whether this plan is feasible.

The solution I proposed above is tuned to the expectations of #1077

sy-records: okay, I got it.

PR #1127, opened 2024-10-10
change minimum username length to 2 instead of 4

### LINKED ISSUE #1124: Remove the minimum username length restriction

I use OAuth to connect accounts. Many users have 3-character usernames in the existing external account system, so is it possible to remove the minimum username length restriction.

Remove the minimum username length restriction.

Set the minimum username length to 2.

zahash:
close #1124


zahash:
@LinkinStars  Ah yes the million i18n strings. Will change those soon

zahash:
changing the i18n strings was very painful because the numbers (4 and 30) are hardcoded. so i had to find all instances that mention those numbers. the annoying part was they were all slightly different `... 4 to 30 ...` | `... 4-30 ...` | `... 4 - 30 ...`

robinv8 (approved):

PR #1135, opened 2024-10-15
feat(pager): Show 404 when page is out of range

### LINKED ISSUE #1133: Improving pagination SEO by handling invalid page requests
## Describe the bug

https://meta.answer.dev/tags/support?page=100

When the requested pagination is greater than the maximum number of paginations, the document should return a 404 status, but now it returns 200

We should add a judgment or directly output 404 status, which is more user-friendly or SEO friendly.

### Expected behavior

When the number of pages exceeds the maximum paging, status should be 404

### Screenshots
image

### Platform

- Device: Desktop
- OS: macOS
- Browser and version: Chrome
- Version: 1.4.0

sy-records:
close #1133

### DIFF
--- a/internal/base/pager/pagination.go
+++ b/internal/base/pager/pagination.go
@@ -62,3 +62,12 @@ func ValPageAndPageSize(page, pageSize int) (int, int) {
 	}
 	return page, pageSize
 }
+
+// ValPageOutOfRange validate page out of range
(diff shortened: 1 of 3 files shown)


shuashuai:
The status of the document should also return 404, so as to improve SEO

now:
image


expected:
image

shuashuai:
@sy-records 

The tag page (/tags) also has this problem

PR #1238, opened 2025-01-29
Answer issue: 1237

sahinakyol:
- Add dependent command 'install-ui-packages' to 'make ui' command
- Add missing old_pass object message to i18n


Makefile
```diff
@@ -47,7 +47,7 @@ install-ui-packages:
 	@corepack enable
 	@corepack prepare pnpm@8.9.2 --activate
 
-ui:
+ui: install-ui-packages
 	@cd ui && pnpm pre-install && pnpm build && cd -
```
shuashuai, line 51: I checked the description in your ISSU. You are actually equivalent to using the local development model. The premise of local development is that you need to install the corresponding `node`、 `pnpm` 、`Golang` environment.
[docs](https://answer.apache.org/docs/development#installation)
image

Makefile does not need to be modified here, because when executing the Makefile completely , will install pnpm first.

sahinakyol: Hello thanks for review.You are right, I'm going to revert this change.

shuashuai (approved):
Thank you, if you are interested in translation, you can continue to contribute through [Crowdin](https://crowdin.com/project/answer) or submitting the code!

PR #1249, opened 2025-02-16
fix: i18n key of ready_desc

gaoxiao6331:
The i18n key is incorrect; it should be ready_desc, so the text is displayed in English.
image


sy-records (approved):
Thanks!

PR #1253, opened 2025-02-17
feat: Add confirm password field in the install form

### LINKED ISSUE #1228: Wrong password during installation makes login impossible
## Is your feature request related to a problem? Please describe

When installing this program, I sometimes accidentally hit wrong keys and enter incorrect passwords. During the installation process, there is no secondary password verification, nor is there an eye icon (👀) that can be clicked to view and confirm the initial password. As a result, if an incorrect password is entered, it becomes impossible to log into the system and the only solution is to reinstall the application.

Add an eye icon or add a second password input field for password confirmation, and check/validate this password.

For now, I haven't thought of any other solutions.

current view:

image

preview demo:

image

CodexX777:
I have added an extra field named confirm password in the application install form to prevent password mismatch. This PR closes #1228

### DIFF
--- a/i18n/en_US.yaml
+++ b/i18n/en_US.yaml
@@ -1692,6 +1692,10 @@ ui:
       msg: Password cannot be empty.
       msg_min_length: Password must be at least 8 characters in length.
       msg_max_length: Password must be at maximum 32 characters in length.
+    admin_confirm_password:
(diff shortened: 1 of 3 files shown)


shuashuai (approved):
Thanks!

PR #1254, opened 2025-02-19
Fix the multi-language translation failure on the installation completion page.

shubo83:
Fix the multi-language translation failure caused by the incorrect i18nKey value on the installation completion page.


sy-records:
Thank you for your PR.

Duplicate of #1249

PR #1256, opened 2025-02-20
feat(build): add plugin vendor directory management function

mylukin:
Implement movePluginToVendor to handle plugin directory organization during build process


mylukin:
I developed a private plugin myself and need UI support


Best Regards,
*Lukin*
Programmer

mylukin:
The original implementation cannot compile my UI into routes because it is limited to the answer-plugins directory, so I need to copy the plugins in other libraries to the answer-plugins directory before compiling.

PR #1257, opened 2025-02-21
feat(ui): improve mobile navigation and tag-based question creation

mylukin:
To improve usability, I’ve placed the 'Add Question' button directly on the main screen, eliminating the need for users to click through a menu. This feature is very important, and positioning it prominently makes it easier for users to find.

image


fenbox:
The "Add Question" button issue has been resolved by the community developers.

The feature of automatically filling in tags when asking a question under a tag might need to be submitted separately.

PR #1259, opened 2025-02-23
fix(install): update language in logged user info store

gaoxiao6331:
When switching the language during installation, the language in the store was not updated, which caused the Accept-Language in the request interceptor to always be set as the default language.
image


ui/src/pages/Install/index.tsx
```diff
@@ -137,6 +140,8 @@ const Index: FC = () => {
     },
   });
 
+  const { update, user } = useStore(loggedUserInfoStore);
+
```
shuashuai, line 144: Instead of using the `useStaore` hook, you can use `loggedUserInfoStore` directly.
eg: 
```
  const { update, user } = loggedUserInfoStore();
```

gaoxiao6331: Ok

shuashuai (approved):
Thanks

PR #1261, opened 2025-02-23
fix(install): update language in logged user info store

gaoxiao6331:
When switching the language during installation, the language in the store was not updated, which caused the Accept-Language in the request interceptor to always be set as the default language. 
image


sy-records:
Thank you for your PR.

You don't need to commit PRs repeatedly, all changes can be made in one PR.

You need to commit to the `dev` branch. Duplicate of #1259

PR #1276, opened 2025-03-10
fix: update admin name validation

### LINKED ISSUE #1273: Name in Admin Account during Installation Should be Username

I noticed that while installing the application, it asks for name and my first guess was to use Full name but it does not support space and it's limited to 30 Characters. Also, this name is used for login. I think the label should be username instead of just name

Name should be change to Username as the description fits more of Username than just name to avoid confusion

Image

sy-records:
Close #1273

### DIFF
--- a/i18n/en_US.yaml
+++ b/i18n/en_US.yaml
@@ -1699,7 +1699,7 @@ ui:
     admin_name:
       label: Name
       msg: Name cannot be empty.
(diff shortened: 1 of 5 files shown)


shuashuai:
> Close #1273

According to @fenbox's  description, the checksum rule here should be wrong, in this case it's the name and not the username!

It should be consistent with the rules for the name on the registration page.

image

shuashuai (approved):
Thanks!

PR #1299, opened 2025-03-28
fix: get tag with page nil pointer error

xuthus5:
When `tc.tagService.GetTagWithPage` exception, resp will get nil value, `resp.XXX` trigger nil pointer error.


sy-records (approved):
Thanks!

PR #1304, opened 2025-04-02
feat(install): auto set the default time zone to where the server is located

### LINKED ISSUE #1303: Auto set the default time zone to where the server is located

After the installation, the default time zone is UTC, which requires one more step to set.

Auto set it to the time zone where the server is located.

hgaol:
Close #1303

### DIFF
--- a/internal/base/constant/constant.go
+++ b/internal/base/constant/constant.go
@@ -29,3 +29,43 @@ var (
 	Revision  = ""
 	GoVersion = ""
 )
+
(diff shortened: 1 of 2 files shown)


hgaol:
No worries, feel free to refactor! I also tried but not found. I think it may because many locations will share the same time offset, if we depends on time offset, we don't know the exact location. If it can get the geo location instead of time offset, maybe can find a better approach.

PR #1314, opened 2025-04-18
fix: Username minimum length still set to 4

### LINKED ISSUE #1313: Username minimum length still set to 4
## Describe the bug

Username minimum length is still set to 4 even though it was supposed to be set to 2 since [v1.4.1](https://github.com/apache/answer/releases/tag/v1.4.1).

### To Reproduce

1. Click my avatar in the top right corner.
2. Click on Settings.
3. In the Profile tab, set Username to some text with 3 letters (e.g., "abc")
4. Click Save button and see error "Username must be greater than 3 characters in length."

### Expected behavior

It should allow usernames with 2 characters and more in length.

### Platform

- Device: Desktop
- OS: Windows 10
- Browser and version: Chrome
- Version: v1.4.5

sy-records:
Fix #1313

### DIFF
--- a/docs/docs.go
+++ b/docs/docs.go
@@ -11050,7 +11050,8 @@ const docTemplate = `{
                 },
                 "username": {
                     "type": "string",
-                    "maxLength": 30
+                    "maxLength": 30,
+                    "minLength": 2
                 },
                 "website": {
                     "type": "string",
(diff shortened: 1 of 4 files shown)


elquimista:
Thanks @sy-records 

I am curious to know what was the culprit behind this issue reoccurring when we already had #1124 sorted out through #1127 - did the relevant lines somehow get overwritten at some point?

sy-records:
It has not been overwritten. #1127 only modifies the restrictions on user registration, but other relevant APIs have not been modified.

PR #1326, opened 2025-04-29
feature: saving files as blob in database

DanielAuerX:
+ files (e.g. branding pictues, attachments) can be saved as a blob in a database. The feature can be turned on and off (feature toggle) with the env var FILE_STORAGE_MODE

+ a new controller with the endpoints /file/branding etc. had to be created since the files are statically linked when fs is used.

+ the files plus metadata are stored in the table 'file'


DanielAuerX:
some clean up tasks etc are needed. just wanted to see if there is interest in a feature like this.

columbo28:
+1 for this PR

I would also like to have that ability as we are using Apache Answers also.

At the moment we have to align the backups of the filesystem AND the database backup as well.
Both of them must match to each other.

If we use the ability to store the files as blobs within the database, we would only have to care about a single backup component.

We could use the toggle that @DanielAuerX already provided to explicitly enable the blob feature, by default we simply set FILE_STORAGE_MODE to 'fs'.

What do you say @LinkinStars ? :)

DanielAuerX:
When I was looking into how files are deleted, I noticed some issues with the file_record table. This is actually related to the problem of keeping the file system and the db in sync.

- Not all uploaded files (branding and avatar) are added to the file_record, leading to the file_record being incomplete
-  If a question (or answer) is deleted, it is not being checked if the text body contained a reference to a file/image. Therefore, the question is deleted, but the file_record entry is not being updated and the image remains in the file system.
- The file record_entity is not registered in init_data.go, therefore the table is not being created in the first place and no files are deleted at all. This also causes an error log whenever the clean up cron job is running.

I might missed or misunderstood some things, but I am quite sure there is something wrong here

DanielAuerX:
Yes sure, I will prepare a PR

sy-records:
Thank you for your contribution.

I also think it would add extra maintenance costs, after all, most use local storage or 3rd party cloud storage. If this can exist as a plugin, that's good.

DanielAuerX:
@mingcheng I can add tests, no problem. However, if I understood @LinkinStars  and @sy-records responses correctly, this pr (feature "saving files as blob" within the answer core) was not gonna be merged?

If its going to be merged, I will continue and finish the pr.

PR #1335, opened 2025-05-09
Fix file record and delete files (branding and avatar)

DanielAuerX:
- [ ] create file_record table
- [ ] avatar and branding files are added to file_record
- [ ] branding files are being deleted
- [ ] avatar files are being deleted


DanielAuerX:
Hi @LinkinStars! yes, the tasks of this pr are finished, unless you see something I missed.
The clean up job (CleanOrphanUploadFiles) is running now. However, I am not sure that it works as intended when it was implemented. Maybe you want to double check that.

ui/src/pages/Users/Settings/Profile/index.tsx
```diff
@@ -241,10 +241,7 @@ const Index: React.FC = () => {
 
     modifyUserInfo(params)
       .then(() => {
-        update({
-          ...user,
-          ...params,
-        });
+        getLoggedUserInfo().then(update);
```
shuashuai, line 244: Why should this be adjusted? 

The whole PR problem has nothing to do with this, and the user information pulled back after editing the user information is also up to date, so there should be no need to adjust here.

DanielAuerX: the browser caches the stale path of the old avatar. Until a new session is started, the old resource is being requested in the backend, causing an error log by the avatar middleware (resource not found).
If you can live with that, ill take it out.

shuashuai: Can you give me a step to reproduce the problem?

I don't know where else will continue to request the old avatar after the user has updated his avatar

DanielAuerX:
@LinkinStars First of all, thanks for the review and the explanation!

> Almost done. I find some small issues.
> 
>     1. Did you forget to submit the `wire gen.go`? We can't run the code directly.
now committed
> 
>     2. If a user **switches** avatars, for example from a customized avatar to the system default avatar, the file **should not be deleted** at this point. This is because the user may switch back later, and the deletion of the old avatar will only be processed when the user **uploads a new custom avatar**.
> 
the avatar file is only being deleted, if the new custom avatar is an empty string or the file name is unlike the old one. Therefore the desired behavior you are describing should already be the case. I also tested it manually. Unless i understand you wrong.

>     3. Let me explain why we use the `CleanOrphanUploadFiles` function to clean the file.
> 
> 
> When a user uploads the file, whatever category(post or avatar), the file will be saved. But after that, if the user does not save the post or profile, the file will be an orphaned file. So we need to clean its background. Of course, we will think that your handling is enough. If you need to be more perfect, it is necessary to separately determine whether the avatar is using it by users(when `isBrandingOrAvatarFile` is true).

you are right, when one uploads an avatar/branding file and never clicks save, the file wont be deleted. I will add some logic to the clean up job, thanks.
I just noticed that the avatar_thumb files arent deleted either. I will also look into that.

DanielAuerX:
> @DanielAuerX
> 
> For issue 2, here are my reproduction steps.
> 
> 1. Upload an avatar and save the profile.
> 2. Check the avatar dir. (The file exists)
> 3. Switch the avatar to system and save the profile.
> 4. Check the avatar dir. (The file not exists)
> 
> ```
> 1. ➜  avatar git:(fix_file_record) ls
> 2. ➜  avatar git:(fix_file_record) ls
> 5sVCXwyhsbm.png
> 4. ➜  avatar git:(fix_file_record) ls
> ➜  avatar git:(fix_file_record) 
> ```
> 
> This is the reason for the problem.
> 
> ```go
> if oldAvatar.Type == "custom" && (newAvatar.Type != "custom" || oldAvatar.Custom != newAvatar.Custom) {
> ```
> 
> At this time, `newAvatar.Type != "custom"` so the file will be deleted.
> 
> To make the logic simple, perhaps it could be implemented like this.
> 
> ```go
> 	// As long as there are no original files, there must be no need to delete
> 	if len(oldAvatar.Custom) == 0 {
> 		return
> 	}
> 	// At this point, it turns out that the old file must exist. 
> 	// Then delete as long as the new file and the old file are inconsistent as you say. 
> 	// Whatever type the user chooses.
> 	if oldAvatar.Custom != newAvatar.Custom {
> 		fileRecord, err := us.fileRecordService.GetFileRecordByURL(ctx, oldAvatar.Custom)
> 	}
> ```

Should be done now

ui/src/pages/Users/Settings/Profile/index.tsx
shuashuai: @DanielAuerX  wait for your reply!

shuashuai: > * checkout branch
> * remove the commit in question
> * upload a custom avatar
> * update the custom avatar with another avatar
> * check the logs and/or the uploads/avatar/ get requests

I tested it and found that this modification still cannot eliminate browser cache unless the entire application is reloaded using `window.reload`, however, using this method will affect the user's experience.

Because from the perspective of generation implementation, updating user information after obtaining from get interface is the same as updating from params, and it cannot eliminate browser cache.

Here are the steps I tested with your code:
1. upload a custom avatar and save
2. reupload a custom avatar and save
3. Return to the question list page, and the last avatar will still be retrieved from the cache (you can see from the console's get request).

shuashuai: > Just for clarification, this change is not supposed to eliminate browser cache. Its supposed to be updating the user information, so an old avatar file is not being requested after it has been removed from the user information. This is not new behavior. However, previously requesting an old resource was successful, because files were not being deleted.

Both ways of updating the user's information are the same and there is no practical difference between them!

> Do you have a better solution? I can also just remove the commit, if you think this is acceptable.

There is no better way at the moment, so let's keep the previous logic for the time being.

PR #1370, opened 2025-07-21
Update Traditional Chinese language files

Oliver0804:
Update Traditional Chinese language files


Oliver0804:
Sorry, I didn't notice that the item hadn't been merged yet when I committed locally and pushed it. I will create a new PR to include only the translation files.






詢問 ChatGPT

Issue #1177, opened 2024-11-14
Make Tags Optional When Creating New Questions


etflly:
## Is your enhancement request related to a problem? Please describe

I'm frustrated when creating new questions because the requirement to add tags can be a barrier. Users may not always know the best tags to use, leading to less accurate categorization.

## Describe the solution you'd like

I would like the option for tags to be optional rather than mandatory when creating new questions. This would allow users to focus on writing their questions and enable administrators to refine tags for accuracy later.

sosyz:
Hi @fenbox, what do you think about this issue? Maybe we can merge it with #842 as a single optimization?

fenbox:
@sosyz Yes, indeed similar, let me think about it.

salimina:
Can I please be assigned this task, and if #842 along with it (due to similar nature)? Thank you!

fenbox:
Add "Min tags per question" in Settings->Write. Default value is `1`. .

Set to `0` means optional, and the "Set required tags" option at the bottom will become disabled and hidden at this time.

<img width="686" alt="image" src="https://github.com/user-attachments/assets/28ffb6fe-e521-474c-bf10-5ebff027ffdb">

PR #1402, opened 2025-10-10
feat(internal,ui): add optional tags

Dinesht04:
fix: #1177 

The video showing working fix:

https://github.com/user-attachments/assets/99f490e3-ffc9-4d8c-a89d-0215932fffcc

Added Migration v28 under the version 1.7.0 for adding a min_tags property and a default value of 1.


LinkinStars:
I tested the previous issue and confirmed it has indeed been resolved.

> I think a good addition could be sending an error to the front-end clearly stating that the tags were not updated due to less than minimum number of tags, or maybe implementing this error and check in the front-end for better UX.
> 

@Dinesht04 You're absolutely right—it would be better if users could receive error messages.

shuashuai (approved):

PR #1403, opened 2025-10-17
feat(internal,ui): add optional question body

### LINKED ISSUE #842: Make "Body" part of the question optional?
via [questions/D1xK2](https://meta.answer.dev/questions/D1xK2/make-body-part-of-the-question-optional)

> I run a community Q&A site where the questions are often very simple such as "Where do I get my dog groomed?". They fit into the Title of the question and there is no need to elaborate on it further in the body. However because the Body is a required field, people are often struggling as to what to put there.

> Can we introduce a feature where the Body field is optional and people can ignore it if they feel the Title captures the question sufficiently?

Dinesht04:
fix #842 

Video showing fix:

https://github.com/user-attachments/assets/d3e634f3-c2d4-4a23-90e6-7dbd64f19601

- Included the addition on min_content in Migration v28 under the version 1.7.0 

Any feedback on this PR would be very welcome

### DIFF
--- a/internal/schema/siteinfo_schema.go
+++ b/internal/schema/siteinfo_schema.go
@@ -75,7 +75,7 @@ type SiteBrandingReq struct {
 
 // SiteWriteReq site write request
 type SiteWriteReq struct {
-	MinimumContent                 int             `validate:"omitempty" json:"min_content"`
+	MinimumContent                 int             `validate:"omitempty,min=0" json:"min_content"`
 	RestrictAnswer                 bool            `validate:"omitempty" json:"restrict_answer"`
 	MinimumTags                    int             `validate:"omitempty" json:"min_tags"`
 	RequiredTag                    bool            `validate:"omitempty" json:"required_tag"`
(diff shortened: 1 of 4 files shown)


ui/src/pages/Questions/Ask/index.tsx
```diff
@@ -423,6 +425,24 @@ const Ask = () => {
...
+    ) {
+      return t(`form.fields.body.hint.optional_body`);
+    }
+
+    let str: string = t(`form.fields.body.hint.minimum_characters`);
+    str = str.replace(
+      `{{ min_content_length }}`,
+      writeInfo.min_content.toString(),
+    );
+    return str;
```
shuashuai: i18next can pass variables, {{}} is the placeholder for the variable

image

internal/schema/siteinfo_schema.go
```diff
@@ -75,6 +75,7 @@ type SiteBrandingReq struct {
 
 // SiteWriteReq site write request
 type SiteWriteReq struct {
+	MinimumContent                 int             `validate:"omitempty" json:"min_content"`
```
LinkinStars: `validate:"omitempty,min=0"` will be better.

ui/src/pages/Questions/Ask/index.tsx
Dinesht04: Thank you for pointing this out, I didn't know that we can pass variables in the i18n function 😅

LinkinStars (approved):

PR #1407, opened 2025-10-23
fix(internal): add greater than and less than equal bounds for min_content and min_body

Dinesht04:
Applied 
 - gt=0,lte=65535 for minimum content (took lte=65535 inference from QuestionAdd in question_schema)
 - gt=0,lte=5 for minimum tags


Dinesht04:
@LinkinStars Thank you for pointing that out. I seemed to overlook it while testing because the validation appears to pass as true even when the value is  = 0  for the gt = 0 tag in the validator. I'll make sure to change it to gte=0 .

Issue #1417, opened 2025-10-29
`<input type="number">` needs to set the `min|max` values on the front-end when necessary


fenbox:
## Is your enhancement request related to a problem? Please describe

* Admin -> Write: When using numeric controls, negative numbers can be entered.
* Admin -> Privileges: The current type is text, and the numeric keyboard cannot be directly called under the touch control system. 

## Describe the solution you'd like

* Set the `min|max` values on the front-end when necessary.
* Use `number` type forms where numbers are involved.

```
<input type="number" min="xxx" max="xxx">
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/number

Dinesht04:
hey, can I work on this?

PR #1424, opened 2025-11-12
feat(ui): add min|max values for inputs and context-based keyboards for inputs (for touch systems)

Dinesht04:
fix: #1417 

Added Property MinValue in Input Component (Default = 0) and Property InputMode (Default = 'text'). Used in making the form for Admin -> Priviliges.

Added InputMode types in InputOptions

Added min = {0} property and inputMode = 'numeric' property to all the numeric type inputs in Admin -> Write.

InputMode hints at the type of data that might be entered by the user while editing the element or its contents. This allows for a numeric keyboard in Touch Mode.
Source for input mode types - https://html.spec.whatwg.org/multipage/interaction.html#attr-inputmode

Didn't add any Default Max values since I think it is subjective and based on user's choice and I couldn't decide an upper limit.
Open to Suggestions.

### DIFF
--- a/ui/src/components/SchemaForm/components/Input.tsx
+++ b/ui/src/components/SchemaForm/components/Input.tsx
@@ -50,10 +50,14 @@ const Index: FC<Props> = ({
   formData,
   readOnly = false,
   min = 0,
-  max = 65355,
+  max,
   inputMode = 'text',
 }) => {
   const fieldObject = formData[fieldName];
+  const numberInputProps =
+    type === 'number'
+      ? { min, ...(max != null && max > 0 ? { max } : {}) }
+      : {};
   const handleChange = (evt: React.ChangeEvent<HTMLInputElement>) => {
     const { name, value } = evt.currentTarget;
     const state = {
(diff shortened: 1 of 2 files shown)


ui/src/components/SchemaForm/components/Input.tsx
```diff
@@ -29,6 +29,17 @@ interface Props {
...
+    | 'text'
+    | 'search'
+    | 'none'
+    | 'tel'
+    | 'url'
+    | 'email'
+    | 'numeric'
+    | 'decimal'
+    | undefined;
 }
```
shuashuai: 1. Just added the relevant TypeScript types inside the input component.
2. Missing relevant TypeScript definitions in SchemaForm(`/ui/src/components/SchemaForm/types.ts`), So there will be a TypeScript error when generating forms through configuration. You can conduct relevant tests in `ui/src/pages/Admin/Smtp/index.tsx`
eg: 
image

3. According to the requirement description, it is also necessary to support the max attribute.

Dinesht04: Thank you for the help. I have added the relevant types, tested the functionality through  SMTP and added the support for max attribute in the latest commit.

PR #1430, opened 2025-11-22
refactor(internal): compile regex once while clearing text

ferhatelmas:
Regex and Replacer can be reused.
No need to recreate for each invocation.


LinkinStars (approved):
LGTM

Issue #1432, opened 2025-11-25
Setup golangci-lint for static analysis of backend


ferhatelmas:
## Is your feature request related to a problem? Please describe

Currently, we lack automated static analysis for Go code, which means code quality issues, bugs, and style inconsistencies are only caught during manual code review or not at all.

## Describe the solution you'd like

Integrate golangci-lint into the repository with a curated set of linters (e.g., errcheck, staticcheck, gosec, govet), a .golangci.yml config file, and CI pipeline integration to run on every PR.

## Describe alternatives you've considered

* Running individual linters manually (fragmented, inconsistent)
* Using go vet alone (limited coverage)
* Relying solely on code review for catching issues (doesn't scale)

ferhatelmas:
@LinkinStars I think we can call this as done and close. Please let me know if anything else is needed otherwise

PR #1445, opened 2025-12-01
refactor(lint): add new linters and fix their issues

ferhatelmas:
* gocritic
* misspell
* modernize (aside, bumping go would be nice)
* testifylint
* unconvert
* unparam
* whitespace

related to #1432

### DIFF
--- a/i18n/cs_CZ.yaml
+++ b/i18n/cs_CZ.yaml
@@ -234,7 +234,7 @@ backend:
         other: Nemáte oprávnění pro aktualizaci.
       content_cannot_empty:
         other: Content cannot be empty.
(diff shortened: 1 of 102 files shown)


i18n/cs_CZ.yaml
```diff
@@ -234,7 +234,7 @@ backend:
         other: Nemáte oprávnění pro aktualizaci.
       content_cannot_empty:
         other: Content cannot be empty.
-      content_less_than_minumum:
+      error.question.content_less_than_minimum:
```
LinkinStars: The replacement here is incorrect.

PR #1449, opened 2025-12-06
feat: add env for glob load template files by gin debug render

liruohrh:
if use that, we can refresh to look new render result of template page but not restart server.


LinkinStars (approved):
LGTM

PR #1453, opened 2025-12-09
refactor: improve queues

ferhatelmas:
* fix race condition for registering handler
* add close method
* use generics to reduce duplication
* rename packages to drop underscore for go convention
* rename interface to drop stutter with package name

### DIFF
--- a/internal/base/queue/queue.go
+++ b/internal/base/queue/queue.go
@@ -26,6 +26,17 @@ import (
 	"github.com/segmentfault/pacman/log"
 )
 
+type Service[T any] interface {
+	// Send enqueues a message to be processed asynchronously.
+	Send(ctx context.Context, msg T)
+
+	// RegisterHandler sets the handler function for processing messages.
+	RegisterHandler(handler func(ctx context.Context, msg T) error)
+
+	// Close gracefully shuts down the queue, waiting for pending messages to be processed.
+	Close()
+}
+
 // Queue is a generic message queue service that processes messages asynchronously.
 // It is thread-safe and supports graceful shutdown.
 type Queue[T any] struct {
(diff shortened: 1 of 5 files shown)


internal/base/queue/queue.go
```diff
@@ -0,0 +1,120 @@
...
+		return
+	}
+
+	select {
+	case q.queue <- msg:
+		log.Debugf("[%s] enqueued message: %+v", q.name, msg)
+	case <-ctx.Done():
+		log.Warnf("[%s] context cancelled while sending message", q.name)
+	}
+}
```
LinkinStars: This function may have a small issue. After determining the closed state, it immediately unlocks the channel and sends a message. If the channel is closed after unlocking, sending the message may result in a “send on closed channel” error. BTW, using only `RLock` here may also pose problems.

internal/service/activityqueue/activity_queue.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"github.com/apache/answer/internal/base/queue"
+	"github.com/apache/answer/internal/schema"
+)
+
+type Service = *queue.Queue[*schema.ActivityMsg]
+
+func NewService() Service {
+	return queue.New[*schema.ActivityMsg]("activity", 128)
+}
```
LinkinStars: We believe defining an interface would be more user-friendly. Such as

```go
type Service interface {
	Send(ctx context.Context, msg *schema.ActivityMsg)
	RegisterHandler(handler func(ctx context.Context, msg *schema.ActivityMsg) error)
	Close()
}

func NewService() Service {
	return queue.New[*schema.ActivityMsg]("activity", 128)
}
```

PR #1458, opened 2025-12-13
feat: publish helm chart

### LINKED ISSUE #1419: Helm Chart Repository invalid
## Describe the bug

The helm Chart Repository pointed out in the https://github.com/apache/answer/blob/main/charts/README.md is wrong.

## To Reproduce

Run

```bash
helm repo add apache https://charts.answer.apache.org/
```

## Expected behavior

Repository to be added, so the helm chart can be installed. 

## Platform

- Device: Desktop
- OS: Ubuntu 24.04
- Version: any

DrFaust92:
Fixes https://github.com/apache/answer/issues/1419

## Proposed Changes

publishes a helm chart so it can be used and no need to fork/duplicate it.

### DIFF
--- a/.github/workflows/helm-release.yaml
+++ b/.github/workflows/helm-release.yaml
@@ -1,0 +1,44 @@
+name: Release Charts
+
+on:
+  push:
+    branches:
+      - oci # change when done testing
+    paths:
+      - 'charts/*'
+
+permissions: read-all
+
+jobs:
+  release:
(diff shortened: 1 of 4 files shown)


LinkinStars:
We don't plan to publish our Helm charts on ghcr.io for the time being. I'm not sure, but I don't think any similar Apache projects have done this before.

PR #1465, opened 2025-12-24
Fix/external id notification

IfDougelseSa:
## Description

  Fixes plugin notifications for SSO-authenticated users by querying the SSO provider's external_id instead of using the notification plugin's name.

  ## Problem

  Plugin notifications were failing for users authenticated via SSO (GitHub OAuth, Google, SAML, etc.) because the code was looking up `external_id` using the notification plugin's name instead of the SSO authentication provider.

  **Example:**
  - User logs in via `provider = "github"` (SSO)
  - Code searches for `provider = "discord"` (notification plugin)
  - No match found → `ReceiverExternalID` is empty

  ## Solution

  Query all external logins for the user and use the most recent one (ORDER BY updated_at DESC).

  ## Changes

  - Add `ORDER BY updated_at DESC` to `GetUserExternalLoginList`
  - Update notification code to query all external logins
  - Use first login from ordered list (most recent)

  ## Testing

  **Before:**
  Receiver External ID:     ← Empty

  **After:**
  Receiver External ID: 81540136     ← Populated

  Tested with GitHub OAuth SSO and test notification plugin.

### DIFF
--- a/i18n/en_US.yaml
+++ b/i18n/en_US.yaml
@@ -2383,6 +2383,7 @@ ui:
     user_normal: This user is already normal.
     user_suspended: This user has been suspended.
     user_deleted: This user has been deleted.
+    user_added: User has been added successfully.
     badge_activated: This badge has been activated.
     badge_inactivated: This badge has been inactivated.
     users_deleted: These users have been deleted.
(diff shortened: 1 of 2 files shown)


LinkinStars:
@IfDougelseSa Originally, this notification plugin was designed to provide consistent notification functionality when users log in with the same account. For example, logging in with Slack would trigger corresponding Slack notifications. Of course, the scenario you mentioned is indeed an issue—users can configure different login methods and notification channels. 

Therefore, our recommendation is to retain both query methods. By default, it will use `externalLogins[0].ExternalID`, but it will also retain the query via `fn.Info().SlugName`. This way, if plugins for the same external system exist, notifications from that specific external system will take precedence.

LinkinStars (changes requested):

PR #1497, opened 2026-02-03
feat: support DeekSeek

MakiWinster72:
add DeekSeek support, now admin can setup DeekSeek in ai-settings, cause most users in China cannt use OpenAI both Authropic and Gemini


MakiWinster72:
image


image

LinkinStars:
According to the Deepseek documentation, it is compatible with OpenAI. Therefore, you only need to select OpenAI and configure `https://api.deepseek.com/v1`.

FYI: https://api-docs.deepseek.com/

MakiWinster72:
> According to the Deepseek documentation, it is compatible with OpenAI. Therefore, you only need to select OpenAI and configure `https://api.deepseek.com/v1`.
> 
> 
> 
> FYI: https://api-docs.deepseek.com/

Thank you! I just found out about this, and I tested it and it worked.

PR #1500, opened 2026-02-08
fix: limit tall image height to improve screenshot viewing experience

MakiWinster72:
small fix

Before
image

After
image

### DIFF
--- a/ui/src/index.scss
+++ b/ui/src/index.scss
@@ -274,6 +274,9 @@ img[src=''] {
   }
   img {
     max-width: 100%;
+    max-height: 70vh;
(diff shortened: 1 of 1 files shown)


fenbox:
This change is not friendly to the scenario of sharing long screenshots.

MakiWinster72:
> This change is not friendly to the scenario of sharing long screenshots.

Yes, since users can click to enlarge the image anyway, right? If screenshots with a high aspect ratio are displayed full width by default, you have to scroll for a long time to reach the bottom, which doesn’t look very nice. 

Of course, this is just a suggestion.

PR #1501, opened 2026-02-10
feat: add prompt configuration for AI settings

MakiWinster72:
I added prompt configuration in AI settings

now admin can edit prompt in "Admin" > "Intelligence" > "AI Settings".

This will display the prompt corresponding to each language in different languages, but currently I have only implemented Chinese and English.

https://github.com/user-attachments/assets/1219cba5-b247-43df-9cff-7f021e59ceb7

### English
image

### Chinese
image

By the way, I also changed the Chinese translation of “intelligence” from “智力” to “智能,” because “智力” is more like the noun “IQ” or “smartness,” used for humans or animals.

### DIFF
--- a/i18n/en_US.yaml
+++ b/i18n/en_US.yaml
@@ -2354,6 +2354,9 @@ ui:
       model:
         label: Model
         msg: Model is required
+      prompt:
+        label: Prompt
+        text: Shows the prompt for the current language. Edit and save to apply.
       add_success: AI settings updated successfully.
     conversations:
       topic: Topic
@@ -2482,5 +2485,4 @@ ui:
     copy: Copy to clipboard
     copied: Copied
     external_content_warning: External images/media are not displayed.
-
 
--- a/i18n/en_US.yaml
+++ b/i18n/en_US.yaml
@@ -2359,6 +2359,9 @@ ui:
         text: Shows the prompt for the current language. Edit and save to apply.
       add_success: AI settings updated successfully.
     conversations:
+      tabs:
+        conversations: Conversations
+        settings: Settings
       topic: Topic
       helpful: Helpful
       unhelpful: Unhelpful
@@ -2485,4 +2488,3 @@ ui:
     copy: Copy to clipboard
     copied: Copied
     external_content_warning: External images/media are not displayed.
-
--- a/ui/src/pages/Admin/AiAssistant/Settings/index.tsx
+++ b/ui/src/pages/Admin/AiAssistant/Settings/index.tsx
@@ -1,0 +1,166 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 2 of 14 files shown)


fenbox:
I think this is a good improvement.

Additionally, the "AI Settings" here define a global AI configuration, not only chat. and the features related to AI Assistant should be moved to "AI Assistant -> Settings".

image

MakiWinster72:
yeah of course.

I’ve moved the prompt configuration from “AI Settings” to “AI Assistant → Settings” as recommended.

humm, I’m newbee to AI, so I’m not entirely sure what other options might eventually belong in the "Settings". I assumed there could be more configurations added in the future, so I made the “Prompt” section collapsible to make it easier to extend later.

image

image

fenbox:
@MakiWinster72 Great, it is also necessary to maintain consistency in the UI, like "Community -> Rules -> Policies"

image

MakiWinster72:
@fenbox Oh, I should have reviewed the other pages more carefully. It should be consistent in style now

image

ui/src/pages/Admin/AiAssistant/index.tsx
```diff
@@ -65,60 +171,117 @@ const Index = () => {
...
+        <Nav.Item>
+          <Nav.Link
+            className="px-0"
+            active={activeTab === 'settings'}
+            onClick={() => setActiveTab('settings')}>
+            {t('tabs.settings')}
+          </Nav.Link>
+        </Nav.Item>
+      </Nav>
+
```
shuashuai, line 79: This is implemented in the way of navigation, not the tabs, you can refer to the line of sight in 'ui/src/pages/Admin/Questions/index.tsx'

image

ui/src/pages/Admin/AiSettings/index.tsx
```diff
@@ -227,6 +227,10 @@ const Index = () => {
       enabled: formData.enabled.value,
       chosen_provider: formData.provider.value,
       ai_providers: newProviders,
+      prompt_config: {
+        zh_cn: historyConfigRef.current?.prompt_config?.zh_cn || '',
+        en_us: historyConfigRef.current?.prompt_config?.en_us || '',
+      },
```
shuashuai: This is related to configuring the global model, prompt_config should be done in a new separate API

MakiWinster72:
Thanks for the correction. At first I only considered Chinese and English, but it should be fine now. pls check at your time.

ui/src/pages/Admin/AiAssistant/Settings/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+  promptConfig: Type.AIPromptConfig | undefined,
+  lang: string,
+) => {
+  if (!promptConfig) {
+    return '';
+  }
+  const isZh = lang?.toLowerCase().startsWith('zh');
+  return isZh ? promptConfig.zh_cn || '' : promptConfig.en_us || '';
+};
+
```
shuashuai, line 40: Normally, the content entered by the user does not need to be translated, is it necessary to load according to the locale? 

@fenbox What about your opinion?

fenbox: @shuashuai The input does not need to be translated, and the chat output will be in the user's set interface language.

shuashuai: @MakiWinster72  Just keep the content filled in by the user, and there is no need to do internationalization

i18n/en_US.yaml
```diff
@@ -2354,8 +2354,14 @@ ui:
...
         label: Model
         msg: Model is required
+      prompt:
+        label: Prompt
+        text: Shows the prompt for the current language. Edit and save to apply.
       add_success: AI settings updated successfully.
     conversations:
+      tabs:
+        conversations: Conversations
+        settings: Settings
```
shuashuai, line 2364: Navigation menus are placed in nav_menus for easy reuse and unified management

Issue #1468, opened 2026-01-05
Add copilot chat plugin to leverage AI to answer user queries based on questions and answers


hgaol:
## Is your feature request related to a problem? Please describe

It's actually a proposal. I think it's very suitable to add AI ability to Apache Answer cause it's an QnA platform. I think we can implement it as a plugin but I'd like to put issue here for more audience. I've already done some POC, feel free to check below demo. Welcome any comments and discussions.

https://github.com/user-attachments/assets/28a848f2-77ee-4dec-9ed3-9dd06c65f21e

## Describe the solution you'd like

As showed in above demo.

There're also some changes needed in answer repo, e.g.
1. supporting some plugin configs depend on others, and only show when it's a specific value.
2. it could be a new plugin type, cause it needs pass the `syncer` and add some REST API for chats.

If it's accepted, I can also implement these as well as this plugin itself.

hgaol:
Hi @fenbox , I've created a PR for this. Instead of my earlier POC (plugin, always-on chat window etc), I use the tools solution which is align with current AI design in Answer, and I think it's consistent for user experience. For more details, pls refer #1510

iabhisekbosepm:
We built essentially this as an external service on top of our production Answer instance (didn't want to wait for a plugin): docs + Q&A → Docling parsing → Qdrant → chat that answers only from the corpus with per-claim citations, reusing Answer login. Running 5 months in prod — ~50% of questions deflected, plus unanswered questions get clustered into a weekly "documentation gaps" report. Live demo: https://project-site-topaz.vercel.app. Happy to share implementation notes if useful for the plugin design.

PR #1510, opened 2026-03-02
feat: support semantic search in AI chat and embedding ability

hgaol:
Implementation of #1468 , add a semantic search tool to keep user experience consistent.

Settings
image

Calling semantic search tool
image

logs
image


internal/repo/embedding/embedding_repo.go
```diff
@@ -0,0 +1,197 @@
...
+	err := r.data.DB.Context(ctx).Find(&list)
+	if err != nil {
+		log.Errorf("get all embeddings failed: %v", err)
+		return nil, err
+	}
+	return list, nil
+}
+
+// SearchSimilar performs brute-force cosine similarity search in Go.
+func (r *embeddingRepo) SearchSimilar(ctx context.Context, queryVector []float32, topK int) ([]SimilarResult, error) {
```
LinkinStars: The biggest issue with this approach lies here: **all data queries are performed in-memory.** While this works for small datasets, it is undoubtedly unacceptable for large-scale data.

Therefore, my suggestion is that if users are using PostgreSQL, they could directly utilize PostgreSQL + pgvector to store vector data and perform searches within the database.

Furthermore, a more 'ideal' approach would be to expose this component as a plugin, allowing user Q&A data to be synchronized to external systems. This design is similar to how search plugins operate. For instance, an Elasticsearch (ES) plugin synchronizes Q&A data to ES for retrieval. The benefit of a plugin-based implementation is extensibility: users aren't restricted to the built-in database and can use their own custom vector databases. The required interfaces would likely mirror those of a search plugin, such as a data synchronization interface and a search interface.

What do you think?

LinkinStars (changes requested):

internal/repo/embedding/embedding_repo.go
hgaol: First of all, thank you @LinkinStars for the careful review! The goal of this PR is also to make it clear, so feel free to raise your concern and suggestions!

> The biggest issue with this approach lies here: all data queries are performed in-memory. While this works for small datasets, it is undoubtedly unacceptable for large-scale data.
>
> Therefore, my suggestion is that if users are using PostgreSQL, they could directly utilize PostgreSQL + pgvector to store vector data and perform searches within the database.

It makes sense. It's better to save vectors in vector DB. And I think it could support multiple vector DBs, while still support saving in main database or memory for test purpose. WDYT?

> Furthermore, a more 'ideal' approach would be to expose this component as a plugin, allowing user Q&A data to be synchronized to external systems. This design is similar to how search plugins operate. For instance, an Elasticsearch (ES) plugin synchronizes Q&A data to ES for retrieval. The benefit of a plugin-based implementation is extensibility: users aren't restricted to the built-in database and can use their own custom vector databases. The required interfaces would likely mirror those of a search plugin, such as a data synchronization interface and a search interface.

About plugin, do you mean make all the functionalities as a plugin, or just the storage for vectors? Per your statement, I assume the later one?

LinkinStars: > And I think it could support multiple vector DBs
> About plugin, do you mean make all the functionalities as a plugin, or just the storage for vectors? Per your statement, I assume the later one?

Yes, if implemented via plugins, it can support different vector databases. I think we can omit the in-memory implementation since it won't be used in most cases.

You're right—it's just a matter of providing **vector storage functionality as a plugin**. Of course, vector search capabilities are also included.

hgaol: Agreed, I love this idea! I'll update the PR per our discussions.

hgaol:
Sorry for late update and I'm kind of busy recently, will start this feature at earliest convenience! Most likely I can start it in next week! Actually it's already WIP.

LinkinStars:
> Sorry for late update and I'm kind of busy recently, will start this feature at earliest convenience! Most likely I can start it in next week! Actually it's already WIP.

@hgaol Don't worry. We have other features in the works as well. So take your time. ♥️

hgaol:
Updated to involve new `vector search` plugin. See demo below.

https://github.com/user-attachments/assets/3b44596e-988d-4576-9a6c-c390177aa5b6

Here's summary of the design and implementation.

---
# Vector Search & Semantic Search Design

## Architecture Layers

```
┌──────────────────────────────────────────────┐
│  AI Chat / MCP Tool ("semantic_search")      │  ← Controller layer
├──────────────────────────────────────────────┤
│  EmbeddingService                            │  ← Service layer (thin facade)
├──────────────────────────────────────────────┤
│  plugin.VectorSearch interface               │  ← Plugin abstraction
├──────────────────────────────────────────────┤
│  pgvector / elasticsearch / weaviate / ...   │  ← Plugin implementations
└──────────────────────────────────────────────┘
```

## Plugin Interface (`plugin/vector_search.go`)

- `RegisterSyncer(ctx, syncer)` — core provides a syncer for bulk data pull
- `SearchSimilar(ctx, query, topK)` — returns `[]VectorSearchResult{ObjectID, ObjectType, Metadata, Score}`
- `UpdateContent(ctx, content)` — upserts a document with embedding
- `DeleteContent(ctx, objectID)` — removes a document
- `ConfigReceiver(config)` / `ConfigFields()` — plugin config lifecycle

`GenerateEmbedding()` is the shared embedding utility used by plugins.

## Content Syncing (`vector_search_sync/syncer.go`)

Core implements `VectorSearchSyncer` with:

- `GetQuestionsPage(page, pageSize)`
- `GetAnswersPage(page, pageSize)`

Each indexed document aggregates question/answer/comment text. Metadata stores deshortened IDs for reconstruction at query time.

Sync is triggered from `RegisterSyncer()` (startup + config update flow).

## Startup & Activation Flow

```
initPluginData():
  1. Load plugin status from DB
  2. Call ConfigReceiver for configured plugins
     -> parse config always
     -> if active: run heavy init (probe embedding + connect/schema checks)
     -> if inactive: skip heavy init (IsEnabled guard)
  3. Call RegisterSyncer for vector search plugins
     -> if active/initialized: trigger full sync
     -> if inactive/uninitialized: skip sync
```

On admin config save:
1. `ConfigReceiver`
2. `UpdatePluginConfig` -> `RegisterSyncer` -> full sync

### Current Behavior Summary

- **Active plugin on startup**: does one probe embedding call, then full sync to vector storage.
- **Inactive plugin on startup**: parses config only; no probe embedding and no sync.
- **Config save for active plugin**: re-runs init path and full sync.

## Semantic Search Query Flow

```
User query -> MCP tool "semantic_search"
  -> EmbeddingService.SearchSimilar(query, topK)
  -> plugin.SearchSimilar() returns scored IDs + metadata
  -> handler fetches full DB content (question/answers/comments)
  -> returns structured semantic search response
```
---

hgaol:
There're also some follow ups I don't have clear answer for now. Post here for discussion.

## Follow-up: When to calc the embeddings and sync to vector storage

### Comparison with Search Plugin

| Aspect | Search Plugin | VectorSearch Plugin |
|--------|---------------|---------------------|
| Bulk sync | Yes | Yes |
| Real-time sync | Yes (create/update/delete hooks) | No |
| Trigger | Event-driven + startup/config | Startup/config only |
| Consistency | Near real-time | Eventually consistent |

### Current Gap

`UpdateContent()` / `DeleteContent()` exist in `plugin.VectorSearch`, but are not called from question/answer service events. So after initial sync, content changes are not reflected until next full re-sync. It's good to keep here since we'll need them when finalized the syncing solution.

### Options

Below are 3 options in my mind. Of course, we can add another setting for each vector plugin to support all 3 options.

1. **Manual (current)**
   - Re-sync only on plugin config save/update
   - Simple, but stale results between syncs

2. **Real-time**
   - Add event hooks to call vector search update/delete
   - Can be async (goroutine / queue) to avoid write-path latency
   - Higher embedding API call volume

3. **Scheduled (cron)**
   - Periodic bulk sync via cron expression
   - Good for off-peak syncing
   - Delayed freshness until next run

hgaol:
I'll create another PR in plugin repo later for memory vector search plugin. I feel that for test purpose, memory has less settings compared to database option. I also implemented some other 3rd party vector store plugins like `weaviate` etc, but some of them haven't validated yet. I'll submit PR for them once done the validations.

hgaol:
Here's the PR for memory vector search plugin. https://github.com/apache/answer-plugins/pull/315

LinkinStars:
> There're also some follow ups I don't have clear answer for now. Post here for discussion.
> 
> ## Follow-up: When to calc the embeddings and sync to vector storage
> ### Comparison with Search Plugin
> Aspect	Search Plugin	VectorSearch Plugin
> Bulk sync	Yes	Yes
> Real-time sync	Yes (create/update/delete hooks)	No
> Trigger	Event-driven + startup/config	Startup/config only
> Consistency	Near real-time	Eventually consistent
> ### Current Gap
> `UpdateContent()` / `DeleteContent()` exist in `plugin.VectorSearch`, but are not called from question/answer service events. So after initial sync, content changes are not reflected until next full re-sync. It's good to keep here since we'll need them when finalized the syncing solution.
> 
> ### Options
> Below are 3 options in my mind. Of course, we can add another setting for each vector plugin to support all 3 options.
> 
> 1. **Manual (current)**
>    
>    * Re-sync only on plugin config save/update
>    * Simple, but stale results between syncs
> 2. **Real-time**
>    
>    * Add event hooks to call vector search update/delete
>    * Can be async (goroutine / queue) to avoid write-path latency
>    * Higher embedding API call volume
> 3. **Scheduled (cron)**
>    
>    * Periodic bulk sync via cron expression
>    * Good for off-peak syncing
>    * Delayed freshness until next run

@hgaol This is a great job—thanks again.

I think your analysis makes sense. We should go with the real-time approach.
  
In my view, semantic search is much more sensitive to stale data than normal keyword search. If we only rely on manual or scheduled full sync, the vector store can easily drift from the source of truth after question/answer/comment edits, status changes, or deletions.
  
So I think the better direction is:
  
  1. keep the existing bulk sync as the bootstrap/rebuild mechanism
  2. add real-time hooks on create/update/delete/status-change events
  3. perform vector store updates asynchronously, so we don’t add latency to the main write path
  
  This way we get both:
  
  - full sync for initialization/recovery
  - near real-time consistency for day-to-day usage

**In short, I haven't seen you use the `DeleteContent` and `UpdateContent` methods.**

I think this will make the design much more complete. We have no issues with the implementation and design of the other interfaces.

hgaol:
> > There're also some follow ups I don't have clear answer for now. Post here for discussion.
> > ## Follow-up: When to calc the embeddings and sync to vector storage
> > ### Comparison with Search Plugin
> > Aspect	Search Plugin	VectorSearch Plugin
> > Bulk sync	Yes	Yes
> > Real-time sync	Yes (create/update/delete hooks)	No
> > Trigger	Event-driven + startup/config	Startup/config only
> > Consistency	Near real-time	Eventually consistent
> > ### Current Gap
> > `UpdateContent()` / `DeleteContent()` exist in `plugin.VectorSearch`, but are not called from question/answer service events. So after initial sync, content changes are not reflected until next full re-sync. It's good to keep here since we'll need them when finalized the syncing solution.
> > ### Options
> > Below are 3 options in my mind. Of course, we can add another setting for each vector plugin to support all 3 options.
> > 
> > 1. **Manual (current)**
> >    
> >    * Re-sync only on plugin config save/update
> >    * Simple, but stale results between syncs
> > 2. **Real-time**
> >    
> >    * Add event hooks to call vector search update/delete
> >    * Can be async (goroutine / queue) to avoid write-path latency
> >    * Higher embedding API call volume
> > 3. **Scheduled (cron)**
> >    
> >    * Periodic bulk sync via cron expression
> >    * Good for off-peak syncing
> >    * Delayed freshness until next run
> 
> @hgaol This is a great job—thanks again.
> 
> I think your analysis makes sense. We should go with the real-time approach.
> 
> In my view, semantic search is much more sensitive to stale data than normal keyword search. If we only rely on manual or scheduled full sync, the vector store can easily drift from the source of truth after question/answer/comment edits, status changes, or deletions.
> 
> So I think the better direction is:
> 
> 1. keep the existing bulk sync as the bootstrap/rebuild mechanism
> 2. add real-time hooks on create/update/delete/status-change events
> 3. perform vector store updates asynchronously, so we don’t add latency to the main write path
> 
> This way we get both:
> 
> * full sync for initialization/recovery
> * near real-time consistency for day-to-day usage
> 
> **In short, I haven't seen you use the `DeleteContent` and `UpdateContent` methods.**
> 
> I think this will make the design much more complete. We have no issues with the implementation and design of the other interfaces.

Thanks @LinkinStars for this suggestion! Agreed with the real-time approach. I'll follow these suggestions and implement it early this week.

hgaol:
- Split strategy:
    - Bulk sync (bootstrap/recovery) via plugin RegisterSyncer: plugin requests pages from core (GetQuestionsPage / GetAnswersPage) and rebuilds embeddings.
    - Realtime sync (steady state) via internal/service/vector_sync: mutation events enqueue async tasks and worker calls UpdateContent / DeleteContent.
  - Core vector flow:
    a. Content mutation occurs (question/answer/comment/review status change).
    b. Service layer enqueues vector task (upsert or delete, object type/id).
    c. Worker resolves active vector plugin (no-op if disabled).
    d. For upsert, core builds VectorSearchContent from canonical data (vector_search_sync builders).
    e. Plugin updates/deletes vector doc.
  - Plugin boundary:
    - Core does orchestration + content shaping.
    - Plugin owns embedding generation/storage/search implementation.
    - Contract remains minimal: UpdateContent, DeleteContent, plus syncer registration for bulk paging.

Below is demo about real-time syncing (async). Thanks!

https://github.com/user-attachments/assets/54ae54ec-1e7b-4132-941d-f272545a1bd1

hgaol:
Hi @LinkinStars , current behavior is as below
1. bulk sync: on App start up, or when saving/updating plugin config(thus no need to restart App to take effect).
2. realtime: CRUD for questions/answers/comments.

I'm thinking to add some checks for bulk syncing, like only updating embeddings for changed content, and skip for unchanged etc. And I plan to do it in another PR to avoid making this PR too big to review.
BTW, I know it's a big PR and need more effort to review. Thanks for your effort and take your time!

LinkinStars (approved):

PR #1511, opened 2026-03-03
feat: add "My" filter

MakiWinster72:
I have added a “My” filter so that users can filter their own questions.
home1
home2


fenbox:
The personal profile lists all of one's own questions, adding here seems redundant.

MakiWinster72:
oh~i see

PR #1512, opened 2026-03-04
refactor(queue): support multiple handlers per queue

william-p:
This PR refactors the internal message queue to support registering multiple handlers per queue instead of a single one.

Previously, calling RegisterHandler would overwrite the existing handler. Now, each call appends to a list of handlers, and all registered handlers are invoked for every message.

_[nb: I am not a Go developer but I have (tried) to intelligently use AI to help me]_

### DIFF
--- a/internal/base/queue/queue.go
+++ b/internal/base/queue/queue.go
@@ -40,12 +40,12 @@ type Service[T any] interface {
 // Queue is a generic message queue service that processes messages asynchronously.
 // It is thread-safe and supports graceful shutdown.
 type Queue[T any] struct {
-	name    string
-	queue   chan T
-	handler func(ctx context.Context, msg T) error
-	mu      sync.RWMutex
-	closed  bool
-	wg      sync.WaitGroup
+	name     string
+	queue    chan T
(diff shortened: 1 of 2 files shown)


LinkinStars:
Using multiple handlers seems like a sound design approach. Perhaps you could elaborate on specific use cases, as currently the entire system typically only registers handlers once—scenarios requiring multiple registrations haven't yet arisen.

william-p:
Today each queue has a single consumer (e.g. only the badge system listens to the event queue), but the event queue is fed by ~15 call sites across the codebase (question/answer/comment CRUD, votes, user updates…). This makes it a natural event bus that additional consumers — audit logging, search reindexing, cache invalidation, analytics — should be able to hook into without replacing the existing handler.

The goal of this PR is to add custom event handler.
Was useful for my Webhook feature (https://github.com/william-p/answer/commits/featWebhooks/ / https://github.com/apache/answer/issues/1509) for catching all events of question/answer (and others ...) life cycle.

The same applies to the notification and activity queues. Generalizing to multi-handler removes this limitation with minimal changes.

PR #1521, opened 2026-04-15
fix: support API key authentication in private mode

### LINKED ISSUE #1508: API Key authentication not supported in private mode (LoginRequired=true)
## Describe the bug

When Apache Answer is configured in private mode (`LoginRequired=true`), API requests authenticated with a valid API Key are rejected with `401 Unauthorized`. The middleware `EjectUserBySiteInfo()` currently only checks for user session cookies and does not verify API Keys, which prevents programmatic access to private instances.

## To Reproduce

1. Configure Apache Answer with `LoginRequired=true` (private mode)
2. Create a valid API Key via Admin Panel → API Keys
3. Make an API request to a public endpoint (e.g., `/answer/api/v1/question/page`) with the API Key in the `Authorization` header:
   ```bash
   curl -H "Authorization: sk_xxxxxxxxxxxxx" \
        "http://localhost:9080/answer/api/v1/question/page?page=1&page_size=10"
   ```
4. Observe the response: `{"code":401,"reason":"base.unauthorized_error","msg":"Unauthorized.","data":null}`

## Expected behavior

API requests with a valid API Key should be allowed to access public endpoints, even when the site is in private mode. API Keys are specifically designed for programmatic access and should bypass the user session requirement.

## Actual Behavior

All API requests without a user session cookie are rejected with `401 Unauthorized`, regardless of whether a valid API Key is provided.

## Root Cause

**File**: `internal/base/middleware/auth.go` (lines 79-108)

The `EjectUserBySiteInfo()` middleware checks if the site is in private mode and then verifies **only** user sessions via `GetUserInfoFromContext(ctx)`. It does not check for API Keys before rejecting the request.

```go
func (am *AuthUserMiddleware) EjectUserBySiteInfo() gin.HandlerFunc {
    return func(ctx *gin.Context) {
        mustLogin := false
        siteInfo, _ := am.siteInfoCommonService.GetSiteSecurity(ctx)
        if siteInfo != nil {
            mustLogin = siteInfo.LoginRequired
        }
        if !mustLogin {
            ctx.Next()
            return
        }

        // If site in private mode, user must login.
        userInfo := GetUserInfoFromContext(ctx)  // ❌ Only checks user session
        if userInfo == nil {
            handler.HandleResponse(ctx, errors.Unauthorized(reason.UnauthorizedError), nil)
            ctx.Abort()
            return
        }
        // ...
    }
}
```

## Impact

This limitation currently prevents some valuable use cases for private Answer instances:
- Automated data synchronization with external systems
- Integration with third-party tools and services
- Headless/API-only clients for programmatic access
- CI/CD pipelines and automated testing

Supporting API Keys in private mode would enable these scenarios while maintaining the security benefits of restricted access.

## Environment

- Apache Answer version: [latest from main branch]
- Configuration: `LoginRequired=true` (private mode)
- API Key scope: `read-only` or `read-write`

## Workaround

Currently, the only workaround is to disable private mode (`LoginRequired=false`), which is not ideal for instances that need to restrict public access.

Herrtian:
## Summary

Closes #1508

When `LoginRequired=true` (private mode), the `EjectUserBySiteInfo` middleware only checked for session-based authentication (cookies). Requests with valid API keys in the `Authorization` header were rejected with 401 Unauthorized.

## Change

**File:** `internal/base/middleware/auth.go` — `EjectUserBySiteInfo()`

When no user session is found in the context, the middleware now falls back to API key validation via `authService.AuthAPIKey()` before rejecting the request. This reuses the same validation logic already used by the `AuthAPIKey()` middleware for MCP routes.

```go
// Before: immediately reject if no user session
if userInfo == nil {
    // 401 Unauthorized
}

// After: try API key before rejecting
if userInfo == nil {
    token := ExtractToken(ctx)
    if len(token) > 0 {
        pass, _ := am.authService.AuthAPIKey(ctx, ctx.Request.Method == "GET", token)
        if pass {
            ctx.Next()
            return
        }
    }
    // 401 Unauthorized
}
```

## Test plan

- [x] `go build ./...` compiles successfully
- [x] Read-only API key + GET request in private mode → allowed
- [x] Write-scope API key + POST request in private mode → allowed
- [x] Read-only API key + POST request in private mode → rejected (scope check)
- [x] No token in private mode → rejected (existing behavior unchanged)
- [x] Session-based auth in private mode → works as before
- [x] Public mode (LoginRequired=false) → no change in behavior

### DIFF
--- a/internal/base/middleware/auth.go
+++ b/internal/base/middleware/auth.go
@@ -92,6 +92,15 @@ func (am *AuthUserMiddleware) EjectUserBySiteInfo() gin.HandlerFunc {
 		// If site in private mode, user must login.
 		userInfo := GetUserInfoFromContext(ctx)
 		if userInfo == nil {
+			// Also check for valid API key authentication.
+			token := ExtractToken(ctx)
+			if len(token) > 0 {
+				pass, _ := am.authService.AuthAPIKey(ctx, ctx.Request.Method == "GET", token)
+				if pass {
+					ctx.Next()
+					return
+				}
+			}
 			handler.HandleResponse(ctx, errors.Unauthorized(reason.UnauthorizedError), nil)
 			ctx.Abort()
 			return


LinkinStars:
Thank you for your contribution. This issue had already been assigned to me and has since been fixed in the dev branch.

PR #1530, opened 2026-05-21
feat: add reasoning content to AI conversation records and update related components

### LINKED ISSUE #1524: deepseek-chat only searches question without retrieving answer, while deepseek-v4-pro / v4-flash cannot respond at all and throw Error: Failed to create AI stream
Describe the bug
deepseek-chat only searches question without retrieving answer, while deepseek-v4-pro / v4-flash cannot respond at all and throw Error: Failed to create AI stream.When the AI assistant is set to deepseek-chat, it can output something, but it only searches for the user's question and cannot effectively retrieve the corresponding answer. Additionally, the system outputs the error message: Error: Failed to create AI stream.

To Reproduce

Set the AI assistant to deepseek-chat mode.

Input a question that requires retrieving relevant information to answer.

Observe that the assistant only echoes or searches for the question itself.

See the error: Error: Failed to create AI stream appearing in the console or response.

Expected behavior
The AI assistant should be able to effectively retrieve the corresponding answer based on the question, instead of only searching for the question text. No Error: Failed to create AI stream should occur.

Screenshots

Image

Image

Image

hgaol:
Fix #1524 

Root cause

DeepSeek's reasoning models stream reasoning_content alongside content. Answer ignored it, so follow-up requests failed with 400: The reasoning_content in the thinking mode must be passed back to the API, and the thinking text was never shown or saved.

Fix

 - Capture reasoning_content from the stream and pass it back to theAPI on subsequent rounds.
 - Persist it with the conversation (new DB column via migrationv2.0.2).
 - Render it in the chat UI as a collapsible "Thinking…/Thoughts"panel above the answer.

Compatibility

Nullable column, omitempty field, UI hides the panel when empty — old conversations and non-reasoning models behave exactly as before.

Demo

https://github.com/user-attachments/assets/49b1a2a1-9133-4ac2-bbeb-860215a50285


LinkinStars (approved):

robinv8 (approved):

LinkinStars:
@hgaol This branch has conflicts that must be resolved. Once the conflicts are resolved, we will merge this PR.

PR #1537, opened 2026-05-28
feat: add recovery middleware to handle panic gracefully

### LINKED ISSUE #1536: The Gin engine is initialized without a recovery middleware, causing connection drops on panic
## Describe the bug

In `internal/base/server/http.go`, the Gin engine is initialized via `gin.New()` without mounting a recovery middleware. When a panic occurs in any handler or middleware, the connection is dropped without returning a proper HTTP response to the client.

## Current behavior

`gin.New()` (unlike `gin.Default()`) does not install any recovery middleware. When a panic propagates out of a handler:

- The Go runtime's `net/http` server recovers per-connection, preventing the process from exiting.
- But it closes the connection without writing any response body.
- The client sees a connection reset / EOF rather than a structured 5xx response.
- The panic is not logged through the project's logging facility (`pacman/log`).

## To Reproduce

This is a defensive coding issue rather than a directly triggerable bug. To verify the behavior, one can temporarily add `panic("test")` to any controller handler and observe that the client receives a connection error instead of an HTTP response.

## Expected behavior

The server should return a unified 500 JSON response with `reason: base.unknown`, consistent with how other errors are handled in the project (see `internal/base/handler/handler.go`). The panic message and stack trace should be logged via the project's logger for diagnostics.

## Additional context

I'd like to submit a PR to fix this by adding a `middleware.Recovery()` that:

- Logs the panic message and full stack trace via `log.Errorf` + `debug.Stack()`.
- Returns the standard error response using `handler.NewRespBody` + `TrMsg` to stay consistent with the rest of the codebase.
- Is mounted as the first middleware in `http.go` so it covers all subsequent middleware and handlers.

hhc7:
Fixes #1536

## Proposed Changes

- Add `middleware.Recovery()` that catches panics from any subsequent middleware or handler, logs the panic message with full stack trace via `log.Errorf` + `debug.Stack()`, and returns a unified 500 JSON response (`reason: base.unknown`) via `handler.NewRespBody` + `TrMsg` —  consistent with how other errors are handled in `internal/base/handler/handler.go`.
 - Mount `Recovery()` as the first middleware in `internal/base/server/http.go` so it covers all subsequent middleware (brotli, accept-language, short-id, auth, etc.) and handlers.
 - Add unit tests in `recovery_test.go` covering both the panic path (verifies 500 +  `base.unknown` reason) and the no-panic path (verifies normal requests pass through unaffected).

### DIFF
--- a/internal/base/middleware/recovery.go
+++ b/internal/base/middleware/recovery.go
@@ -22,6 +22,7 @@ package middleware
 import (
 	"net/http"
 	"runtime/debug"
+	"strings"
 
 	"github.com/apache/answer/internal/base/handler"
 	"github.com/apache/answer/internal/base/reason"
@@ -29,14 +30,31 @@ import (
 	"github.com/segmentfault/pacman/log"
 )
 
-func Recovery() gin.HandlerFunc {
+func Recovery(apiPrefixes ...string) gin.HandlerFunc {
 	return func(ctx *gin.Context) {
 		defer func() {
 			if err := recover(); err != nil {
 				log.Errorf("panic recovered: %v\n%s", err, debug.Stack())
-				ctx.AbortWithStatusJSON(http.StatusInternalServerError,
-					handler.NewRespBody(http.StatusInternalServerError, reason.UnknownError).TrMsg(handler.GetLangByCtx(ctx)),
-				)
+
+				// Headers/body already flushed (SSE or any streamed response).
+				// We can no longer rewrite the response cleanly; just stop the chain.
+				if ctx.Writer.Written() {
+					ctx.Abort()
+					return
+				}
+
+				path := ctx.Request.URL.Path
+				for _, p := range apiPrefixes {
+					if strings.HasPrefix(path, p) {
+						ctx.AbortWithStatusJSON(http.StatusInternalServerError,
+							handler.NewRespBody(http.StatusInternalServerError, reason.UnknownError).
+								TrMsg(handler.GetLangByCtx(ctx)),
+						)
+						return
+					}
+				}
+
+				ctx.AbortWithStatus(http.StatusInternalServerError)
 			}
 		}()
 		ctx.Next()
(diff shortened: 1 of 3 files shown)


LinkinStars:
I agree that adding a recovery layer makes sense. gin.New() does not include one by default, so today a panic can indeed end up as a dropped connection instead of a controlled response.

  That said, I do not think this implementation is quite correct yet, because it assumes every panic can be turned into the same JSON 500 response.

  There are at least two cases where that is not very elegant:

  1. SSE / streaming endpoints
     For /chat/completions, the handler starts an SSE response very early: it sets Content-Type: text/event-stream, sends status 200, flushes headers, and may already write streamed chunks before later logic runs. If a panic happens after that point,
     recovery cannot reliably convert the response into a JSON 500. In practice it may just append JSON into an already-started event stream, which produces an invalid/mixed response rather than a clean error.

  2. HTML routes
     This server also handles UI / HTML routes, not only JSON APIs. With a global recovery mounted at the engine level, a panic in an HTML route would now also return the API-style JSON body. That is probably not the intended behavior for page requests,
     and it is not really “consistent” for the whole server, only for JSON endpoints.

  So I think the underlying concern is valid, but the current fix is a bit too broad. A more robust approach would be to either:

  - scope JSON recovery only to API routes, or
  - make recovery response-aware, for example checking whether headers/body were already written and whether the request is an API, SSE, or HTML route.

  In short: the direction is reasonable, but handling all panics with the same AbortWithStatusJSON(500, ...) is not especially elegant for SSE and HTML paths.

hhc7:
Thanks for the thoughtful review. I'll rework the middleware along the response-aware approach (your second suggestion):

1. Check ctx.Writer.Written() first. If true (SSE / already-flushed responses), only log + abort, never try to write a body.
2. Otherwise, branch by route family — JSON 500 for API routes (\*/api/\*), and AbortWithStatus(500) for HTML routes so the browser shows its native error page rather than raw JSON.

Keeping recovery mounted at the engine level guarantees no future route group accidentally loses panic coverage. Will push the update shortly.

LinkinStars (approved):

PR #1540, opened 2026-06-04
feat(site): allow disabling email verification

### LINKED ISSUE #1517: [Feature] Add option to disable email verification on registration
### Problem
Currently, Apache Answer **always requires** email verification after registration. Users cannot access the platform until they click the verification link in their email.

This is problematic because:
- Many users don't receive or check the verification email
- It creates friction in user onboarding
- Some deployments (internal networks, dev environments) don't need email verification

### Proposed Solution
Add an admin configuration option to allow users to use the platform immediately after registration without email verification.

#### 1. Schema Changes
**File:** `internal/schema/siteinfo_schema.go`
```go
type SiteLoginReq struct {
    AllowNewRegistrations    bool     `json:"allow_new_registrations"`
    AllowEmailRegistrations bool     `json:"allow_email_registrations"`
    AllowPasswordLogin      bool     `json:"allow_password_login"`
    AllowEmailDomains       []string `json:"allow_email_domains"`
    RequireEmailVerification bool    `json:"require_email_verification"` // NEW
}
```

#### 2. Registration Logic
**File:** `internal/service/content/user_service.go`

Current behavior (line ~471):
```go
userInfo.MailStatus = entity.EmailStatusToBeVerified  // Always unverified
```

New behavior:
```go
if requireEmailVerification {
    userInfo.MailStatus = entity.EmailStatusToBeVerified
} else {
    userInfo.MailStatus = entity.EmailStatusAvailable  // Verified immediately
}
```

#### 3. Admin API
**File:** `internal/controller_admin/siteinfo_controller.go`
- Update `SiteLoginReq` struct to include new field
- Ensure admin can toggle this setting

#### 4. Frontend Admin Panel
**File:** `ui/src/admin/pages/site/` (or similar)
- Add toggle switch for "Require Email Verification" in login settings

### Default Behavior
- `RequireEmailVerification = true` (maintain backward compatibility)
- Admin can set to `false` to allow immediate access after registration

### Database Changes
None required - existing `mail_status` field already supports both states:
- `1` = EmailStatusAvailable (verified)
- `2` = EmailStatusToBeVerified (pending)

### Testing Checklist
- [ ] New user registration with verification enabled → mail_status = 2
- [ ] New user registration with verification disabled → mail_status = 1
- [ ] Admin can toggle the setting via admin panel
- [ ] API returns correct value for `require_email_verification`
- [ ] Existing users with mail_status = 2 still need to verify (if required)

### Related Files
- `internal/schema/siteinfo_schema.go`
- `internal/service/content/user_service.go`
- `internal/controller_admin/siteinfo_controller.go`
- `internal/service/siteinfo_common/siteinfo_service.go`
- `ui/src/` (admin panel components)

AsyncAssassin:
Fixes #1517

## Proposed Changes

  - Add a `require_email_verification` login setting that defaults to `true` to preserve the current behavior.
  - Expose the setting through the admin login settings API and UI.
  - Allow newly registered email users to become active immediately when email verification is disabled.
  - Keep existing unverified users unchanged.
  - Add migration and regression tests for legacy login settings and registration behavior.

## Testing

  - `go test ./internal/migrations -run 'TestBackfillRequireEmailVerification|TestAddRequireEmailVerification|TestSplitLegalMenuKeepsRequireEmailVerificationDefaultTrue' -count=1 -v`
  - `go test ./...`
  - `make lint`
  - `make ui`
  - `make build`


LinkinStars (changes requested):
The compatibility and default-safe behavior in this change look solid, and I agree with the overall direction. That said, I think there are still a few backend design / maintainability issues worth revisiting before merge:

1. OptionalBool feels too heavy for this use case. To handle the omitted / null / false cases for require_email_verification, the PR introduces a dedicated OptionalBool type in the schema layer. The issue is that this API does not read like a clearly defined patch-style endpoint; it looks more like a normal save/update API. Adding a special tri-state wrapper for a single field increases the cognitive load in both the schema and save logic.
If this is intended to be a full save API, I would prefer requiring the client to send the full field set. If this is intended to be a partial update API, then it would be better to make that explicit and handle partial-update semantics consistently rather than special-casing one field.

2. Treating null as true is safe, but not very intuitive semantically.
In the current implementation, both the migration path and the runtime save path normalize null to true. I understand why this was done from a security/default-safety perspective, but from an API semantics perspective, null usually reads more like “unset” than “explicitly enabled”. If we want to keep this behavior, I think it should at least be documented more explicitly in code/comments; otherwise it is not obvious to future maintainers why null does not preserve the current value, fail validation, or behave like an omitted field.

3. applyEmailVerificationSetting is a very thin abstraction. This helper currently does little more than req.SkipEmailVerification = !siteInfo.RequireEmailVerification. Extracting that into a separate function adds another jump for the reader without really encapsulating meaningful logic yet. Unless we expect more registration policy logic to live there soon, I think keeping this mapping inline at the call site would actually be clearer.

4. The naming mixes a positive configuration flag with a negative execution flag. At the config level we have RequireEmailVerification, while at the request/service level we introduce SkipEmailVerification, which means we have to invert the meaning in between. This works, but it hurts readability and makes future changes easier to get wrong. I would strongly prefer keeping the naming direction consistent across layers, for example using RequireEmailVerification end-to-end, or using a single enabled/disabled wording consistently.

AsyncAssassin:
Thanks for the detailed review. I updated the login settings save endpoint to work as a full-save API instead of adding partial-update semantics for this one field.

I removed OptionalBool and made require_email_verification required in the save request. Null or omitted values are now rejected by validation and by SaveSiteLogin. For existing stored configs, the safe legacy default is still preserved: missing/null values default to true, while an explicit false stays false. I also added a short comment in the v34 migration to make that default clear.

I removed applyEmailVerificationSetting and renamed the internal registration flag from SkipEmailVerification to RequireEmailVerification, so the naming stays positive and consistent through the registration flow. The registration fallback behavior is still covered by tests.

I regenerated Swagger so require_email_verification is marked as a required request field.

Validation:
- go test ./internal/...
- git diff --check

LinkinStars (approved):
LGTM

robinv8 (approved):

sy-records (approved):

PR #1543, opened 2026-06-09
i18n: complete Russian (ru_RU) translation

neermilov:
### What does this PR do?

Completes the Russian (`ru_RU`) localization. All strings that were still in English in `i18n/ru_RU.yaml` are now translated, and the language progress in `i18n/i18n.yaml` is bumped from 80% to 100%.

Alse refined a few already-translated strings so the text fits the UI better without changing its meaning.

### Why translate here instead of Crowdin?

The Crowdin source is currently out of date and does not contain all of the keys present in the repository. The files in the repo are the up-to-date source of truth, so the translation was done directly here to cover every key.

### Verification

- Key parity with `en_US.yaml`: 1597/1597 keys present.
- `ru_RU.yaml` parses as valid YAML.

### DIFF
--- a/i18n/i18n.yaml
+++ b/i18n/i18n.yaml
@@ -43,7 +43,7 @@ language_options:
     progress: 96
   - label: "Русский"
     value: "ru_RU"
-    progress: 80
+    progress: 100
   - label: "简体中文"
     value: "zh_CN"
     progress: 100
--- a/i18n/ru_RU.yaml
+++ b/i18n/ru_RU.yaml
@@ -140,7 +140,7 @@ backend:
   pass:
     other: Пароль
   old_pass:
-    other: Current password
+    other: Текущий пароль
   original_text:
     other: Это сообщение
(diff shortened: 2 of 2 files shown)


Snehajit999:
Looks good to me. Thanks for completing the Russian localization and ensuring full key parity with `en_US.yaml`. The explanation for translating directly in the repository instead of Crowdin also makes sense given the outdated source there.

haosb:
@LinkinStars Hello, could you review this please? This blocking us to use it at production, we've thought about fork approaches, but it doesn't look good in the future.

LinkinStars (approved):

sy-records (approved):
Maybe we should update Crowdin Source.

LinkinStars:
> @LinkinStars Hello, could you review this please? This blocking us to use it at production, we've thought about fork approaches, but it doesn't look good in the future.

@haosb Thanks for the reminder. This will be included in the next release.

PR #1544, opened 2026-06-09
fix: accept answer fails when short links enabled (#1541)

### LINKED ISSUE #1541: Can't accept answers
## Describe the bug

After a recent update, I am no longer able to accept answers. I always get:

## To Reproduce

Happens on any answer I try to accept. Click the "Accept" button and be presented with `Error... Answer do not found.` dialog

## Expected behavior

Clicking "Accept" marks an answer as accepted.

## Platform

- Device: Desktop
- OS: macOS
- Browser and version: Zen 1.20.2b, but happens in all browsers
- Version: 2.0.1

hgaol:
Close #1541 

The ownership check added to AcceptAnswer compared the answer's QuestionID against the request's QuestionID directly. When short links are enabled, answerRepo.GetByID re-encodes QuestionID to its short form while the controller de-shorts req.QuestionID to its long form, so the two encodings of the same question never matched and every accept returned "Answer do not found". Normalize both ids via uid.DeShortID before comparing, preserving the privilege-escalation guard for answers that truly belong to another question.

Validated in local environment.

### DIFF
--- a/internal/service/content/answer_service.go
+++ b/internal/service/content/answer_service.go
@@ -471,7 +471,7 @@ func (as *AnswerService) AcceptAnswer(ctx context.Context, req *schema.AcceptAns
 		}
 
 		// check answer belong to question
-		if acceptedAnswerInfo.QuestionID != req.QuestionID {
+		if !sameObjectID(acceptedAnswerInfo.QuestionID, req.QuestionID) {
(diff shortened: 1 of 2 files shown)


LinkinStars (approved):
Thank you for your contribution. This fix is very important.

Issue #625, opened 2023-11-15
more options for email notifications


chenfengrugao:
1. 现在的邮件通知方式是新帖子全员通知，会造成一瞬间发送大量email的情况，邮件服务商似乎没法全部成功处理，是否可以增加一个每封邮件之间的发送间隔的设置
2. 提问回复可以立即发送邮件通知，这个没问题。新帖子和关注标签，是否可以改成一天发送一次
3. 如果用户处于登录状态，站内提醒就可以。建议此时不发送email通知，以免用户email收到大量通知类邮件。
谢谢

mysterious-cod1004:
Can I resolve it now ?

AsyncAssassin:
Hi, I’d like to work on the first item here.

Based on the comment from @LinkinStars, I’ll keep the scope limited to adding a configurable interval between emails sent for new-question / followed-tag notification fan-out. I won’t change daily digest behavior or logged-in user handling in this PR.

My current plan is to use an env var with a default that preserves the current behavior, and apply the delay only to these notification emails, not to verification, password reset, or SMTP test emails.

LinkinStars:
> Hi, I’d like to work on the first item here.
> 
> Based on the comment from [@LinkinStars](https://github.com/LinkinStars), I’ll keep the scope limited to adding a configurable interval between emails sent for new-question / followed-tag notification fan-out. I won’t change daily digest behavior or logged-in user handling in this PR.
> 
> My current plan is to use an env var with a default that preserves the current behavior, and apply the delay only to these notification emails, not to verification, password reset, or SMTP test emails.

@AsyncAssassin That's a good plan.

PR #1545, opened 2026-06-17
feat(notification): add interval for new question emails

AsyncAssassin:
## Summary

Adds a configurable interval between email send attempts for new-question / followed-tag notification fan-out.

The interval is controlled by `NEW_QUESTION_NOTIFICATION_EMAIL_SEND_INTERVAL_SECONDS`.

The default is `0`, so the current behavior is preserved unless the environment variable is explicitly set. Empty, invalid, negative, or overflowing values also fall back to `0`.

## Scope

This is based on the scope discussed in #625 and confirmed by @LinkinStars.

This PR only applies the delay to new-question notification email fan-out. It does not change:

- daily digest behavior
- logged-in user handling
- verification emails
- password reset emails
- SMTP test emails
- the shared `EmailService.Send`
- UI/admin settings
- migrations

When the interval is enabled, plugin notifications are sent before the throttled email fan-out, so they are not delayed by the configured sleep.

## Tests

Added focused tests for:

- environment variable parsing
- delay ordering between email attempts
- disabled/non-email channel handling
- unsubscribe token generation per email attempt
- service-level new-question fan-out behavior with fake service dependencies

Validation:

```bash
go test -count=1 ./internal/service/notification/...
go test -race -count=1 ./internal/service/notification/...
go test -count=1 ./internal/...
go test -count=1 ./...
go vet ./internal/service/notification/...
git diff --check
```

### DIFF
--- a/internal/service/notification/external_notification.go
+++ b/internal/service/notification/external_notification.go
@@ -45,6 +45,7 @@ type ExternalNotificationService struct {
 	notificationQueueService   noticequeue.ExternalService
 	userExternalLoginRepo      user_external_login.UserExternalLoginRepo
 	siteInfoService            siteinfo_common.SiteInfoCommonService
+	newQuestionEmailWorker     *newQuestionEmailWorker
 }
 
 func NewExternalNotificationService(
@@ -67,6 +68,10 @@ func NewExternalNotificationService(
 		userExternalLoginRepo:      userExternalLoginRepo,
 		siteInfoService:            siteInfoService,
 	}
+	n.newQuestionEmailWorker = newQuestionEmailWorkerWithDefaults(
+		newQuestionNotificationEmailSendInterval,
+		n.sendNewQuestionNotificationEmail,
+	)
 	notificationQueueService.RegisterHandler(n.Handler)
 	return n
 }
--- a/internal/service/notification/new_question_email_worker.go
+++ b/internal/service/notification/new_question_email_worker.go
@@ -1,0 +1,308 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package notification
+
+import (
+	"context"
+	"sync"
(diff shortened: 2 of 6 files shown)


LinkinStars:
Thanks for working on this. The scope matches the discussion in #625, and the default `0` interval preserves the current behavior.

One concern: the current implementation sleeps inside `handleNewQuestionNotification`. Since `external_notification` is processed by a single queue worker, this does not only throttle new-question emails. A large fan-out with a configured interval can block the whole external notification queue, delaying unrelated notifications such as new answer, new comment, and invite-answer emails.

There is also a shutdown concern because the wait uses `time.Sleep`, so it cannot be cancelled. If the interval is large or misconfigured, queue shutdown may have to wait for the sleep to finish.

I think a better approach would be to move the throttled new-question email fan-out into a dedicated queue/worker:

- `handleNewQuestionNotification` collects subscribers, syncs plugin notifications, enqueues new-question email tasks, and returns immediately.
- A dedicated `new_question_email` worker consumes those tasks serially.
- The worker waits between email attempts using a context-aware timer/select instead of bare `time.Sleep`.
- The interval should probably have a reasonable upper bound to avoid accidental long blocking from misconfiguration.

This would keep the intended email throttling behavior while avoiding delays for unrelated external notifications and making shutdown behavior safer.

AsyncAssassin:
Thanks for the review. I pushed a follow-up commit that moves the throttled new-question email fan-out out of `handleNewQuestionNotification` and into a dedicated serial `new_question_email` worker.

The handler now collects subscribers, syncs plugin notifications, enqueues the email task, and returns without sleeping or sending throttled SMTP inline. The worker uses a bounded queue, context-aware timer/select waits, and clamps the configured interval to a 5-minute maximum. Unrelated external notifications should no longer be delayed by new-question email throttling.

I also added coverage for the worker timing/cancellation behavior, enqueue/close race, full/closed queue cases, handler enqueue behavior, plugin-before-enqueue ordering, and interval parsing.

LinkinStars:
Thanks for the follow-up update. The new dedicated worker design addresses the main concern about blocking unrelated external notifications, and the context-aware wait plus interval cap look good. I think this is very close now; there is just one remaining concern about the worker queue size.

Right now `newQuestionEmailWorkerQueueSize` is fixed at `128`, and when the queue is full, `TryEnqueue` drops the whole new-question email task. For sites with many followed tags/subscribers, SMTP sending can be much slower than new-question creation, so this queue may fill up quickly and later notification emails would be skipped.

Could we change the default queue size from `128` to `1024`, and also make it configurable by environment variable? For example:

- default: `1024`
- env var: `NEW_QUESTION_NOTIFICATION_EMAIL_QUEUE_SIZE`
- invalid, empty, zero, or negative values fall back to `1024`
- optionally clamp extremely large values to a reasonable maximum

This should cover most installations by default, while still allowing larger sites to tune the backlog capacity based on their user count and SMTP throughput.

AsyncAssassin:
Thanks, I pushed a small follow-up commit for the worker queue size.

The `new_question_email` worker now defaults to a queue size of `1024` and can be configured with `NEW_QUESTION_NOTIFICATION_EMAIL_QUEUE_SIZE`. Invalid, empty, zero, negative, and overflow values fall back to `1024`, and very large values are clamped to `65536`.

I also added focused tests for the parser and constructor capacity.

LinkinStars (approved):

PR #1548, opened 2026-06-29
fix: add proper HTML structure to email templates to reduce spam score

### LINKED ISSUE #783: Spam Analysis report of mail triggered. High spam score of TVD_PH_BODY_META_ALL of 2.7
## Describe the bug

Usually mail provider consider 5 and above mails are spam.

when I ran spam analysis on triggered mails. There are the results

image

Spam Analysis is performed by Apache Spam Assasin https://spamassassin.apache.org/

TVD_PH_BODY_META_ALL  this Absence of missing header and body make spam detectors to consider it as SPAM mail

### Screenshots

If applicable, add screenshots or video to help explain your problem.

### Platform

- Device: [e.g. Desktop, Mobile]
- OS: [e.g. macOS]
- Browser and version: [e.g. Chrome, Safari]
- Version: [e.g. v1.2.0]

Yash-Chauhan22:
## Summary

Fixes #783

Emails sent by Answer were being flagged by spam filters (Apache SpamAssassin) 
with the `TVD_PH_BODY_META_ALL` rule due to missing proper HTML document 
structure in the email body.

## Root Cause

The `TVD_PH_BODY_META_ALL` rule fires when an HTML email lacks:
- `<!DOCTYPE html>` declaration
- `<html>`, `<head>`, `<body>` tags
- `<meta charset>` tag

## Changes

- **`internal/service/export/email_service.go`**: Added `MIME-Version: 1.0` 
  header and explicit `charset=UTF-8` in `Content-Type` for all outgoing emails
- **`i18n/en_US.yaml`**: Wrapped all 8 email body templates with proper HTML 
  document structure (`<!DOCTYPE html>`, `<html>`, `<head>`, `<meta charset='UTF-8'>`, 
  `<meta name='viewport'>`, `<body>`)

## Testing

Verified Go build passes with no errors. Fix can be validated using 
[mail-tester.com](https://www.mail-tester.com) or Apache SpamAssassin locally.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,5 +5,13 @@
   "explorer.autoReveal": "focusNoScroll",
   "cSpell.words": [
     "grecaptcha"
-  ]
+  ],
+  "go.goroot": "${env:USERPROFILE}\\tools\\go",
+  "go.gopath": "${env:USERPROFILE}\\go",
+  "terminal.integrated.env.windows": {
+    "PATH": "${env:USERPROFILE}\\tools\\go\\bin;${env:USERPROFILE}\\tools\\node-v22.16.0-win-x64;${env:PATH}",
+    "GOROOT": "${env:USERPROFILE}\\tools\\go",
+    "GOPATH": "${env:USERPROFILE}\\go"
+  },
+  "go.toolsManagement.autoUpdate": true
 }
--- a/GSOC.md
+++ b/GSOC.md
@@ -1,0 +1,95 @@
+﻿# GSoC 2026 — Apache Answer Contribution Log
+
+> **Contributor:** Yash Chauhan (@Yash-Chauhan22)
+> **Upstream Project:** [apache/answer](https://github.com/apache/answer)
+> **Fork:** [Yash-Chauhan22/answer](https://github.com/Yash-Chauhan22/answer)
+> **Program:** Google Summer of Code 2026
+
+---
+
+## Repository Setup
+
+| Task | Status |
+|------|--------|
+| Fork apache/answer | Done |
+| Clone fork locally | Done |
+| Add upstream remote | Done |
+| Install Go toolchain | In progress |
+| Install Node.js + pnpm | In progress |
+| Create .env config | Done |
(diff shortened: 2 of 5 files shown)


LinkinStars:
Thanks for working on this. The overall direction makes sense to me. Using a complete HTML document structure for HTML emails is a common practice and should help reduce compatibility issues with mail clients and spam filters.

  However, I think this PR needs a few adjustments before it can be merged.

  1. Please remove unrelated changes from this PR.

  The following files/changes are not related to the email spam-score fix and should be excluded:

  - .vscode/settings.json
  - GSOC.md
  - docker-compose.yaml
  - unrelated Go doc comment updates in email_service.go

  This PR should focus only on the email template / email sending behavior.

  2. Please revert the MIME changes in email_service.go.

  gomail.NewMessage() already uses UTF-8 by default, and SetBody("text/html", body) already writes the HTML part with Content-Type: text/html; charset=UTF-8.

  gomail also automatically writes Mime-Version: 1.0 if it is not already present.

  The current change may cause duplicate or malformed headers:

  m.SetHeader("MIME-Version", "1.0")

  m.SetBody("text/html; charset=UTF-8", body)

  In gomail, the auto-check uses the exact header key Mime-Version, while this PR sets MIME-Version, so gomail may still add its own Mime-Version header. Also, gomail appends ; charset=UTF-8 when writing the body part, so passing text/html; charset=UTF-8 to SetBody may result in duplicated charset
  parameters.

  So I suggest keeping the original call:

  m.SetBody("text/html", body)

  3. Please apply the HTML document structure consistently.

  Currently only i18n/en_US.yaml is updated. Since email templates are translated based on the user language, other locales will still send partial HTML fragments without DOCTYPE, html, head, meta, and body.

  If we keep the HTML document structure inside i18n templates, then the email templates across locales should be updated consistently, or at least all supported email template bodies should use the same HTML document wrapper.

  In short, I agree with the main fix, but the PR should be narrowed down to the email template changes, avoid manual MIME header changes, and remove unrelated development-environment/doc changes.

PR #1563, opened 2026-07-26
fix: tag search never matches on slug name

ngineer420:
## Proposed Changes

Tag search never matches on `slug_name`.

In `tagCommonRepo.GetTagPage`, the search term is formatted into `LOWER(%s)` and
passed as the **value** of the `LIKE`, so the function name ends up inside the
pattern rather than being applied to the column:

```go
builder.Like{"slug_name", fmt.Sprintf("LOWER(%s)", tag.SlugName)}
// -> slug_name LIKE '%LOWER(coco)%'
```

That can never match. Only the `display_name` clause does any work, and `LIKE`
is case-sensitive on PostgreSQL, so searching for a tag by the name it is
written in returns nothing:

```
GET /answer/api/v1/tags/page?slug_name=Coco   ->  ["coco"]
GET /answer/api/v1/tags/page?slug_name=coco   ->  []
```

Tags are lower case by convention, so lower case is what people type, and the
tag filter appears to report that no such tag exists.

  - Lower both the column and the term so the comparison is case-insensitive
  - Extract the term normalisation into `searchTermForTag` so the behaviour can
    be covered by a unit test without standing up a database
  - Add that test

Reproduced against 2.0.2 on PostgreSQL 18.

### DIFF
--- a/internal/repo/tag_common/tag_common_repo.go
+++ b/internal/repo/tag_common/tag_common_repo.go
@@ -21,7 +21,6 @@ package tag_common
 
 import (
 	"context"
-	"fmt"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


LinkinStars (approved):
LGTM

PR #1564, opened 2026-07-26
feat: add a banner area below the site header

ngineer420:
## Proposed Changes

Answer has a **Header** area in Admin → CSS and HTML. The client writes its content at the top of `<body>`:

```js
const handleCustomHeader = (content) => {
  const el = document.body;
  renderCustomArea(el, CUSTOM_MARK_HEADER, 'afterbegin', content);
};
```

The server template writes it in the same place, before `<div id="root">`.

That position is above the nav bar and outside `#root`. The nav bar and the page are both inside `#root`, and the injected node is a sibling of `#root`. No stylesheet can move it below the nav. A banner under the site header is not possible today.

This PR **adds a new Banner area**. It does not move the Header area.

- `custom_banner` is a new field on `SiteCustomCssHTMLReq`, beside `custom_header` and `custom_sidebar`.
- `pages/Layout` renders `<CustomBanner />` below `<Header />`.
- Admin → CSS and HTML gets a **Banner** field, beside Head, Header, Footer and Sidebar.

`ui/template/header.html`, `internal/controller/template_controller.go` and `ui/src/components/Customize/index.tsx` are unchanged. The diff adds 136 lines and removes none.

## Answers to the review

**1. Breaking change.** Removed. The Header area keeps its position and its behaviour, so existing sites see no change. The banner is a separate optional field, and it stays empty until an admin fills it in. `CustomBanner` renders nothing while the field is empty.

**2. Scripts must run.** `CustomBanner` re-creates the script nodes after it renders the slot, so the browser runs them. It runs them one time for each value, and it tracks the value it activated. One difference from `ActivateScriptNodes`: it leaves a data block alone, such as a script with type `application/ld+json`. The browser never runs a data block, and the IIFE wrapper would corrupt its content.

**3. Head and footer refactor.** Reverted. `ui/src/components/Customize/index.tsx` matches `main` byte for byte, and the `meta[name="go-template"]` skip stays.

## Why the server template does not paint the banner

The first version of this PR painted the banner in `header.html` and read that markup back in the client. That cannot work, and it would have caused the same double execution described in point 3 above.

`ui/src/index.tsx` calls `ReactDOM.createRoot`, not `hydrateRoot`. `App` returns `<InitialLoadingPlaceholder />` on its first render, because `useMergeRoutes` starts with an empty route list and fills it in an effect. That placeholder is the first commit of the root, and `createRoot` clears the container on the first commit. So React empties `#root` before `Layout` renders at all. Any banner the server painted inside `#root` is already gone, and the browser has already run its scripts once while it parsed the document.

The banner is therefore client-rendered only, in the same way as `custom_sidebar` and `CustomSidebar`. `#spin-mask` covers the viewport until React mounts, so a reader never sees the difference.

## Checks

- `go build ./...` and `go vet` pass.
- `tsc --noEmit`, `eslint` and `prettier --check` pass on the changed files.
- Swagger docs regenerated with `swag` v1.16.3. The ASF licence headers on `docs/docs.go` and `docs/swagger.yaml` are kept.

### DIFF
--- a/docs/docs.go
+++ b/docs/docs.go
@@ -11864,6 +11864,10 @@ const docTemplate = `{
         "schema.SiteCustomCssHTMLReq": {
             "type": "object",
             "properties": {
+                "custom_banner": {
+                    "type": "string",
+                    "maxLength": 65536
+                },
                 "custom_css": {
                     "type": "string",
                     "maxLength": 65536
@@ -11889,6 +11893,10 @@ const docTemplate = `{
         "schema.SiteCustomCssHTMLResp": {
             "type": "object",
             "properties": {
+                "custom_banner": {
+                    "type": "string",
+                    "maxLength": 65536
+                },
                 "custom_css": {
                     "type": "string",
                     "maxLength": 65536
--- a/docs/swagger.json
+++ b/docs/swagger.json
@@ -11837,6 +11837,10 @@
         "schema.SiteCustomCssHTMLReq": {
             "type": "object",
             "properties": {
+                "custom_banner": {
+                    "type": "string",
+                    "maxLength": 65536
+                },
                 "custom_css": {
                     "type": "string",
                     "maxLength": 65536
@@ -11862,6 +11866,10 @@
         "schema.SiteCustomCssHTMLResp": {
             "type": "object",
             "properties": {
+                "custom_banner": {
+                    "type": "string",
+                    "maxLength": 65536
+                },
                 "custom_css": {
                     "type": "string",
                     "maxLength": 65536
--- a/docs/swagger.yaml
+++ b/docs/swagger.yaml
@@ -2353,6 +2353,9 @@ definitions:
     type: object
   schema.SiteCustomCssHTMLReq:
     properties:
+      custom_banner:
+        maxLength: 65536
+        type: string
       custom_css:
         maxLength: 65536
         type: string
@@ -2371,6 +2374,9 @@ definitions:
     type: object
   schema.SiteCustomCssHTMLResp:
     properties:
+      custom_banner:
+        maxLength: 65536
+        type: string
       custom_css:
         maxLength: 65536
         type: string
--- a/i18n/en_US.yaml
+++ b/i18n/en_US.yaml
@@ -2225,6 +2225,9 @@ ui:
       header:
         label: Header
         text: This will insert after &lt;body>
+      banner:
+        label: Banner
+        text: This will insert below the site header.
       footer:
         label: Footer
         text: This will insert before &lt;/body>.
--- a/internal/schema/siteinfo_schema.go
+++ b/internal/schema/siteinfo_schema.go
@@ -239,6 +239,7 @@ type SiteCustomCssHTMLReq struct {
 	CustomHead    string `validate:"omitempty,gt=0,lte=65536" json:"custom_head"`
 	CustomCss     string `validate:"omitempty,gt=0,lte=65536" json:"custom_css"`
 	CustomHeader  string `validate:"omitempty,gt=0,lte=65536" json:"custom_header"`
+	CustomBanner  string `validate:"omitempty,gt=0,lte=65536" json:"custom_banner"`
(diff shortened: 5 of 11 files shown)


robinv8 (changes requested):
Thanks for the PR — placing a banner below the site nav is a reasonable goal, and rendering it inside the React tree to survive hydration is the right approach. But after testing, I have to request changes:

1. **This is a breaking change, not just a fix.** `custom_header` moves from the top of `<body>` to below the nav. Existing sites rely on the current position (top announcement bars, scripts that must run before `#root`). Please call this out explicitly in the PR description — or better, keep the current position and add a separate slot below the nav so existing users are unaffected.

2. **Scripts in the custom header no longer execute.** The old path goes through `ActivateScriptNodes`, which re-creates `<script>` nodes so they run. `dangerouslySetInnerHTML` never executes scripts, so any header content containing JS silently breaks. Please activate script nodes after rendering the slot.

3. **Please revert the head/footer refactor.** The `go-template` skip is not redundant: app settings are loaded before render (`guard.ts:431`), so on SSR pages the store is already populated at mount. The new unconditional injection clears the server-rendered markup and `ActivateScriptNodes` executes every script a second time (analytics double-counting, etc.). Keeping the skip avoids this.

Happy to re-review once these are addressed.

ngineer420:
@robinv8 Thanks for the detailed review. All three points are addressed, and the approach changed.

The PR no longer moves `custom_header`. It adds a separate `custom_banner` field and renders it below the site header. `ui/src/components/Customize/index.tsx`, `ui/template/header.html` and `internal/controller/template_controller.go` are now unchanged from `main`. The diff adds 130 lines and removes none.

1. **Breaking change** — gone. The Header area keeps its position and its behaviour. The banner is a separate optional field, and it renders nothing until an admin fills it in.
2. **Scripts** — `CustomBanner` re-creates the script nodes after it renders the slot, and runs them one time for each value. It wraps only a classic script, so a data block such as `application/ld+json` keeps its content.
3. **Head and footer refactor** — reverted. `Customize/index.tsx` matches `main` byte for byte, and the `meta[name="go-template"]` skip stays.

Your third point also showed that the first version was wrong in a second way. `App` returns `<InitialLoadingPlaceholder />` on its first render, because `useMergeRoutes` starts with an empty route list. That placeholder is the first commit of the root, and `createRoot` clears the container on the first commit. So React empties `#root` before `Layout` renders at all. A banner painted into `#root` by the server cannot survive that, and the browser has already run its scripts once during parse. The banner is therefore client-rendered only, in the same way as `custom_sidebar` and `CustomSidebar`.

robinv8:
Thanks for rewriting this around a separate `custom_banner` — that does address the three points from the earlier review (no breaking move of `custom_header`, scripts run, and `Customize` / the Go template stay untouched).

After trying it locally, though, I’m not convinced we need a new Admin field for this. `custom_header` already runs scripts, and a small script can wait for `#header` and insert markup right under the nav. I verified that on a local instance.

Minimal example to paste into **Admin → CSS and HTML → Header**:

```html
<script>
(function () {
  var ID = 'hello-under-header';
  function mount() {
    var header = document.getElementById('header');
    if (!header || document.getElementById(ID)) return;
    var el = document.createElement('div');
    el.id = ID;
    el.textContent = 'Hello World — under the site header';
    el.style.cssText = 'padding:8px 16px;text-align:center;background:#e7f1ff;';
    header.insertAdjacentElement('afterend', el);
  }
  mount();
  var root = document.getElementById('root') || document.body;
  new MutationObserver(mount).observe(root, { childList: true, subtree: true });
})();
</script>
```

So I’d rather keep the surface area small: document that Header-script pattern for “banner below the nav”, and not add `custom_banner`. Happy to help with a short docs note if useful. For this PR, I’d suggest closing it (or leaving it unmerged) unless there’s a reason a first-class field is still worth the maintenance.

Appreciate the careful write-up either way.

fenbox:
This feature is suitable as a plugin and is not recommended to be included in the core program.

PR #1565, opened 2026-08-02
test(converter): pin renderLinkIsUrl behavior

ferhatelmas:
Add a table-driven test covering the markdown link destination check
before replacing the govalidator dependency with stdlib logic.


mingcheng (approved):
lgtm

LinkinStars (approved):

PR #1566, opened 2026-08-03

xantorres:

The configuration preserves the two output contracts the server depends on:
the build directory stays at `ui/build`, which `ui/static.go` embeds, and
emitted assets stay under `static/`, which `internal/router/ui.go` serves as a
route. Files keep the `static/js`, `static/css` and `static/media` grouping,
because both `.gitignore` and the `analyze` script match on that layout.
Production sourcemaps stay on, and the `REACT_APP_` prefix is retained so
`ui/scripts/env.js` remains the single source of truth for configuration
shared with the server. That same script's `public_url` value now also
drives Vite's `base` configuration, and the manifest link in `ui/index.html`
uses Vite's `%BASE_URL%` macro instead of a hardcoded path, so a non-root
deployment keeps every asset and manifest reference prefixed the way it did
under the previous toolchain's `PUBLIC_URL` wiring.

Agentic tooling did the mechanical work in this migration; every change was
reviewed by a human before being committed.

### Server-side changes, and why they were unavoidable

Three parts of the Go side had one bundler's output format encoded into them.
A frontend-only migration was attempted first and is not possible.

**`GetStyle()` matched a literal tag shape.** It scraped `index.html` with
regexes requiring classic scripts with `defer` first and stylesheet links with
`href` before `rel`. Any other shape returned nothing, and server-rendered
pages would load with no JavaScript and no stylesheet while every build step
still reported success. The tags are now read from the parsed document, so
attribute order, attribute set and quoting no longer matter, and the next
bundler change cannot reintroduce this.

**`header.html` re-emitted those paths as classic scripts.** An ES module
loaded through a classic script tag fails on its first import, so fixing the
parsing alone would still have shipped broken server-rendered pages. The tag is
now declared as a module. Rewriting the built `index.html` instead was not an
option, because `internal/router/ui.go` serves that same file to boot the SPA
and it cannot misdeclare its own script type.

**`GetStyle()` returned a single stylesheet.** The previous build emitted
exactly one entry stylesheet; the current one emits two, so pages loaded
partially unstyled. It is now a list, mirroring how script paths were already
collected and prefixed.

### Route code splitting

Routes loaded pages with `` lazy(() => import(`@/pages/${pagePath}`)) ``. That
shape cannot be statically analyzed, so no page received its own chunk and the
specifier reached the browser untransformed, leaving every lazily routed page
unable to load. Pages are now enumerated with a bounded glob covering the three
directory depths routes actually use, excluding component subtrees so their own
index files do not become route chunks.

A page path with no matching module now rejects with the requested path and the
list of known keys, surfacing through the existing route error boundary.

### Three behaviour changes, called out deliberately

**The markdown editor now renders its themed background.**
`src/components/Editor/index.scss` read `var(-bs-body-bg)` with a single leading
dash. That is not a valid custom property reference, so the declaration was
discarded and the editor never received the background it asks for. The
previous CSS minifier accepted the invalid value; the current one rejects it,
which is how it surfaced. Fixing it changes rendering.

**The custom stylesheet link now derives its href from the build's own base,
not a separately computed value.** `PageTags` built the `/custom.css` link
from `process.env.PUBLIC_URL`, which the previous toolchain exposed with its
trailing slash already stripped; at the default configuration that value was
an empty string, resolving to `/custom.css`. `import.meta.env.BASE_URL`, the
direct equivalent under the new toolchain, keeps the trailing slash, so
substituting it in the same place would resolve the same default
configuration to `//custom.css` instead. The trailing slash is stripped
explicitly before the substitution, reproducing the previous output at the
default configuration and staying correct away from it. The output here is
unchanged; only the mechanism it depends on is.

**Two routes were dead ends; one now renders, and the other now fails loudly
instead of silently.** `pages/403` and `pages/Admin/UserOverview` both
referenced modules that did not exist in the tree, and both failed the same
way, silently rendering blank. `pages/403` has been repointed at
`pages/404/403`, an existing component, and now renders correctly; that
route is no longer dead. `pages/Admin/UserOverview` still references a
module that does not exist; it now reports the missing path through the
route error boundary instead of rendering blank. Neither is a regression:
the `pages/403` fix restores a working page, and the
`pages/Admin/UserOverview` change only makes an existing gap visible.
Creating that missing page file is a separate, content decision.

### Dependency removals

`react-scripts`, `react-app-rewired`, `customize-cra` and `config-overrides.js`
are gone, and `yaml-loader` is replaced by the equivalent Vite plugin, pinned
to the schema the previous loader used so bare dates and merge keys keep
parsing the same way they did before.

Three removed packages were already inert before this migration. Both purgecss
packages were declared but wired nowhere: no postcss config exists, the
overrides file never referenced them, and no script invoked them. `buffer` was
aliased and provided as a global, but no application source uses it, and the
one dependency requiring it declares `buffer: false` in its own browser field.

`sass` and `@types/node` are raised to the versions the toolchain requires; the
previous `sass` predates the async compiler API it now calls. `sass` is pinned
below the release that begins deprecating `@import`, which this project uses
across 30 files. Migrating those to `@use` is a separate concern. Bootstrap's
own Sass internals print dozens of dependency deprecation warnings on every
build, unrelated to anything in this project's own styles; `vite.config.mts`
sets `css.preprocessorOptions.scss.quietDeps: true` to silence those
specifically while still surfacing warnings from this project's own
stylesheets.

The eslint config no longer extends `react-app/jest`, which shipped inside
`react-scripts` and configured rules for a test suite this project does not
have.

### A third failure, found by running the built application

With the build green, the two checks above passing and the dev server working,
the built application did not boot. React never mounted, the page showed its
loading spinner indefinitely, and the browser console was empty.

`i18next` attaches its resource-store methods to the instance inside `init()`.
The builtin plugins register their translations while their modules are being
evaluated. Whether that happens before or after `init` depends on how the
bundler groups and orders chunks, so the previously working order was
incidental rather than guaranteed. When it inverts, the registration throws
while the entry module is still evaluating, which takes the application down
before it mounts and produces no console output.

Registration now happens immediately only when there is an initialised instance
to register into, and otherwise falls to the `initialized` handler the code
already installed, which is correct in either order.

### Review hardening

A closer pass after the initial migration found several places where the new
toolchain worked but was not yet equivalent to the old one.

**Bootstrap icon fonts.** The bootstrap-icons stylesheet points at font files
under a path the bundler could not resolve on the first migration pass, so
the build silently emitted zero font files and every icon rendered as a
missing-glyph box. The stylesheet now overrides the package's font-directory
variable to a path Vite can resolve; the fonts are emitted, and the browser
boot test described below confirms they are served.

**Type checking.** `pnpm build` now also runs `tsc --noEmit` before the
bundler runs, and is clean today. The previous toolchain ran its checker in
the dev server (blocking) and during builds (downgraded to warnings by this
project's `TSC_COMPILE_ON_ERROR` setting); none of that carried over when the
bundler changed, so until this fix a type error shipped with no signal at all.

**Yaml parsing.** The Vite yaml plugin defaults to js-yaml's more permissive
schema, which resolves bare dates to JS `Date` objects and enables merge
keys; the previous loader's schema kept both as plain strings. The plugin is
now pinned to that same schema.

**Module script tag shape.** `header.html`'s module script tag now carries
`crossorigin`, matching the tag Vite's own build emits for the equivalent
client-rendered entry point. This does not create a new requirement: a
`type=module` script already fetches in CORS mode, so any CDN origin serving
these files already needs to send the matching CORS headers, with or without
`crossorigin` present.

**Declared Node range.** The bundler's declared support range is
`^20.19.0 || >=22.12.0`. `ui/package.json`'s own `engines.node` allowed
`>=20`, which admits versions below that floor, and now matches it exactly.
The release workflow's pinned Node, which also sat below the floor, is raised
to `20.19.0` to match. That version bump is the only change this branch makes
anywhere under `.github/`; it does not add a job.

**Guard checks now fail closed.** All three `make check-ui` checks preset
their result to failure and only report success once they have proved it.
The two Node scripts, `check-locale-resolution.js` and
`check-plugin-i18n-order.js`, bound every server-creation, listen and close
call with its own timeout and add an unreferenced whole-run watchdog, so a
hang cannot silently drain the event loop into an accidental success instead
of a reported failure. `check-built-assets.sh` gained a preflight that fails
if the Go test it depends on is ever renamed or removed, plus a fourth
self-check fixture confirming a non-stylesheet `<link rel="manifest">` is
never miscounted as a stylesheet.

**Typed environment variables.** `import.meta.env.REACT_APP_*` previously
typed as `any`, since no `ImportMetaEnv` augmentation existed; a typo'd key
would compile clean and only surface as a missing value at runtime. The two
keys the app reads this way, `REACT_APP_API_URL` and `REACT_APP_BASE_URL`,
are now declared, with `strictImportMetaEnv` enabled so an undeclared key is
a type error instead of a silent `any`.

### Verification

- `pnpm build` completes with no errors and exactly two warnings:
  `front-matter`, a dependency unrelated to this project's own pinned
  `js-yaml@^4.1.0`, pulls in a legacy `js-yaml@3.x` copy whose `buffer`
  import gets externalized for browser compatibility, and one chunk exceeds
  the default size threshold; see the note on chunking below. `tsc --noEmit`,
  which now runs as part of this build, is clean.
- `make check-ui` passes, covering three behaviours a successful build does
  not demonstrate: that the server can still find the built asset paths,
  that a non-default language resolves and is fetchable at runtime, and that
  plugin translations register regardless of module evaluation order,
  verified through both the deferred and the immediate registration paths.
  All three checks fail closed: each presets its result to failure, bounds
  every step that can hang with its own timeout, and only reports success
  once it has proved it.
- Every check was observed failing before being relied on.
  `./script/check-built-assets.sh --self-check` rewrites the built
  `index.html` with four shapes the server cannot parse, including a
  non-stylesheet manifest link, and asserts the check fails on each, so a
  check that quietly stopped asserting anything is distinguishable from a
  passing one. The same script also fails if the Go test it depends on is
  ever renamed or removed, instead of silently matching nothing.
- On the Go side: `go build ./...` succeeds, so the embed still resolves;
  `go vet ./...` reports no issues; `TestGetStyleResolvesBuiltAssets` passes;
  and the license header check (`script/check-asf-header.sh`) passes.
- The built application was booted against sqlite3 and loaded in a browser.
  `/`, `/questions`, `/tags` and `/users` render with the client mounted,
  every asset and font loads, and the console is empty.

This project has no test runner and no unit tests, before or after. These are
three targeted regression checks, not a test suite.

### Measurements

Same machine, same Node and package manager versions, clean tree and clean
install on both sides, five runs per timing metric.

| Metric | Before | After |
|---|---|---|

- Cold and warm builds are within noise of each other, because the new build
  has no meaningful persistent cache to warm and now also runs a type check
  on every build, cold or warm alike. The old toolchain had a real warm
  cache, which is why its two figures differ so much. The comparison to draw
  is cold against cold.
- Dev server time-to-ready is wall clock on both sides, including process
  spawn. The new tool self-reports a much smaller number that excludes that,
  and using it would compare two different quantities.
- Chunk boundaries differ structurally between the two bundlers, so the
  bundle rows compare total shipped bytes rather than like-for-like chunks.
- HMR latency was measured once, at commit 7200ca3, on both toolchains, and
  carried forward from there: none of the commits since touch the hot-update
  path.
- All figures were captured at the branch head, d06b623, with a clean
  working tree on both sides.

### Not included

This PR adds no CI job for the frontend. The project runs no frontend job
today, and a build on every push is a cost a maintainer should choose to
take on, not one this migration should impose. One constraint carries
forward for whoever wires that job later: a bare `go test ./...` reports ok
while asserting nothing about the built asset paths, because
`TestGetStyleResolvesBuiltAssets` skips when no frontend build is embedded.
Any future CI job needs to build the frontend first.

The dev server now binds to loopback only by default; reaching it from
another device on the network needs an explicit `--host` flag.

Create React App's SVG-as-component imports
(`import { ReactComponent as X } from './x.svg'`) are not carried over to
this configuration. Nothing in this codebase used them.


xantorres:
Superseded by #1567. Renaming the head branch closed this PR; the branch and commits are unchanged. Apologies for the noise.

PR #1567, opened 2026-08-03

xantorres:

The configuration preserves the two output contracts the server depends on: the build directory stays at `ui/build`, which `ui/static.go` embeds, and emitted assets stay under `static/`, which `internal/router/ui.go` serves as a route. Files keep the `static/js`, `static/css` and `static/media` grouping, because both `.gitignore` and the `analyze` script match on that layout. Production sourcemaps stay on, and the `REACT_APP_` prefix is retained so `ui/scripts/env.js` remains the single source of truth for configuration shared with the server. That same script's `public_url` value now also drives Vite's `base` configuration, and the manifest link in `ui/index.html` uses Vite's `%BASE_URL%` macro instead of a hardcoded path, so a non-root deployment keeps every asset and manifest reference prefixed the way it did under the previous toolchain's `PUBLIC_URL` wiring.

Agentic tooling did the mechanical work in this migration; every change was reviewed by a human before being committed.

### Server-side changes, and why they were unavoidable

Three parts of the Go side had one bundler's output format encoded into them. A frontend-only migration was attempted first and is not possible.

**`GetStyle()` matched a literal tag shape.** It scraped `index.html` with regexes requiring classic scripts with `defer` first and stylesheet links with `href` before `rel`. Any other shape returned nothing, and server-rendered pages would load with no JavaScript and no stylesheet while every build step still reported success. The tags are now read from the parsed document, so attribute order, attribute set and quoting no longer matter, and the next bundler change cannot reintroduce this.

**`header.html` re-emitted those paths as classic scripts.** An ES module loaded through a classic script tag fails on its first import, so fixing the parsing alone would still have shipped broken server-rendered pages. The tag is now declared as a module. Rewriting the built `index.html` instead was not an option, because `internal/router/ui.go` serves that same file to boot the SPA and it cannot misdeclare its own script type.

**`GetStyle()` returned a single stylesheet.** The previous build emitted exactly one entry stylesheet; the current one emits two, so pages loaded partially unstyled. It is now a list, mirroring how script paths were already collected and prefixed.

### Route code splitting

Routes loaded pages with `` lazy(() => import(`@/pages/${pagePath}`)) ``. That shape cannot be statically analyzed, so no page received its own chunk and the specifier reached the browser untransformed, leaving every lazily routed page unable to load. Pages are now enumerated with a bounded glob covering the three directory depths routes actually use, excluding component subtrees so their own index files do not become route chunks.

A page path with no matching module now rejects with the requested path and the list of known keys, surfacing through the existing route error boundary.

### Three behaviour changes, called out deliberately

**The markdown editor now renders its themed background.** `src/components/Editor/index.scss` read `var(-bs-body-bg)` with a single leading dash. That is not a valid custom property reference, so the declaration was discarded and the editor never received the background it asks for. The previous CSS minifier accepted the invalid value; the current one rejects it, which is how it surfaced. Fixing it changes rendering.

**The custom stylesheet link now derives its href from the build's own base, not a separately computed value.** `PageTags` built the `/custom.css` link from `process.env.PUBLIC_URL`, which the previous toolchain exposed with its trailing slash already stripped; at the default configuration that value was an empty string, resolving to `/custom.css`. `import.meta.env.BASE_URL`, the direct equivalent under the new toolchain, keeps the trailing slash, so substituting it in the same place would resolve the same default configuration to `//custom.css` instead. The trailing slash is stripped explicitly before the substitution, reproducing the previous output at the default configuration and staying correct away from it. The output here is unchanged; only the mechanism it depends on is.

**Two routes were dead ends; one now renders, and the other now fails loudly instead of silently.** `pages/403` and `pages/Admin/UserOverview` both referenced modules that did not exist in the tree, and both failed the same way, silently rendering blank. `pages/403` has been repointed at `pages/404/403`, an existing component, and now renders correctly; that route is no longer dead. `pages/Admin/UserOverview` still references a module that does not exist; it now reports the missing path through the route error boundary instead of rendering blank. Neither is a regression: the `pages/403` fix restores a working page, and the `pages/Admin/UserOverview` change only makes an existing gap visible. Creating that missing page file is a separate, content decision.

### Dependency removals

`react-scripts`, `react-app-rewired`, `customize-cra` and `config-overrides.js` are gone, and `yaml-loader` is replaced by the equivalent Vite plugin, pinned to the schema the previous loader used so bare dates and merge keys keep parsing the same way they did before.

Three removed packages were already inert before this migration. Both purgecss packages were declared but wired nowhere: no postcss config exists, the overrides file never referenced them, and no script invoked them. `buffer` was aliased and provided as a global, but no application source uses it, and the one dependency requiring it declares `buffer: false` in its own browser field.

`sass` and `@types/node` are raised to the versions the toolchain requires; the previous `sass` predates the async compiler API it now calls. `sass` is pinned below the release that begins deprecating `@import`, which this project uses across 30 files. Migrating those to `@use` is a separate concern. Bootstrap's own Sass internals print dozens of dependency deprecation warnings on every build, unrelated to anything in this project's own styles; `vite.config.mts` sets `css.preprocessorOptions.scss.quietDeps: true` to silence those specifically while still surfacing warnings from this project's own stylesheets.

The eslint config no longer extends `react-app/jest`, which shipped inside `react-scripts` and configured rules for a test suite this project does not have.

### A third failure, found by running the built application

With the build green, the two checks above passing and the dev server working, the built application did not boot. React never mounted, the page showed its loading spinner indefinitely, and the browser console was empty.

`i18next` attaches its resource-store methods to the instance inside `init()`. The builtin plugins register their translations while their modules are being evaluated. Whether that happens before or after `init` depends on how the bundler groups and orders chunks, so the previously working order was incidental rather than guaranteed. When it inverts, the registration throws while the entry module is still evaluating, which takes the application down before it mounts and produces no console output.

Registration now happens immediately only when there is an initialised instance to register into, and otherwise falls to the `initialized` handler the code already installed, which is correct in either order.

### Review hardening

A closer pass after the initial migration found several places where the new toolchain worked but was not yet equivalent to the old one.

**Bootstrap icon fonts.** The bootstrap-icons stylesheet points at font files under a path the bundler could not resolve on the first migration pass, so the build silently emitted zero font files and every icon rendered as a missing-glyph box. The stylesheet now overrides the package's font-directory variable to a path Vite can resolve; the fonts are emitted, and the browser boot test described below confirms they are served.

**Type checking.** `pnpm build` now also runs `tsc --noEmit` before the bundler runs, and is clean today. The previous toolchain ran its checker in the dev server (blocking) and during builds (downgraded to warnings by this project's `TSC_COMPILE_ON_ERROR` setting); none of that carried over when the bundler changed, so until this fix a type error shipped with no signal at all.

**Yaml parsing.** The Vite yaml plugin defaults to js-yaml's more permissive schema, which resolves bare dates to JS `Date` objects and enables merge keys; the previous loader's schema kept both as plain strings. The plugin is now pinned to that same schema.

**Module script tag shape.** `header.html`'s module script tag now carries `crossorigin`, matching the tag Vite's own build emits for the equivalent client-rendered entry point. This does not create a new requirement: a `type=module` script already fetches in CORS mode, so any CDN origin serving these files already needs to send the matching CORS headers, with or without `crossorigin` present.

**Declared Node range.** The bundler's declared support range is `^20.19.0 || >=22.12.0`. `ui/package.json`'s own `engines.node` allowed `>=20`, which admits versions below that floor, and now matches it exactly. The release workflow's pinned Node, which also sat below the floor, is raised to `20.19.0` to match. That version bump is the only change this branch makes anywhere under `.github/`; it does not add a job.

**Guard checks now fail closed.** All three `make check-ui` checks preset their result to failure and only report success once they have proved it. The two Node scripts, `check-locale-resolution.js` and `check-plugin-i18n-order.js`, bound every server-creation, listen and close call with its own timeout and add an unreferenced whole-run watchdog, so a hang cannot silently drain the event loop into an accidental success instead of a reported failure. `check-built-assets.sh` gained a preflight that fails if the Go test it depends on is ever renamed or removed, plus a fourth self-check fixture confirming a non-stylesheet `<link rel="manifest">` is never miscounted as a stylesheet.

**Typed environment variables.** `import.meta.env.REACT_APP_*` previously typed as `any`, since no `ImportMetaEnv` augmentation existed; a typo'd key would compile clean and only surface as a missing value at runtime. The two keys the app reads this way, `REACT_APP_API_URL` and `REACT_APP_BASE_URL`, are now declared, with `strictImportMetaEnv` enabled so an undeclared key is a type error instead of a silent `any`.

### Verification

- `pnpm build` completes with no errors and exactly two warnings: `front-matter`, a dependency unrelated to this project's own pinned `js-yaml@^4.1.0`, pulls in a legacy `js-yaml@3.x` copy whose `buffer` import gets externalized for browser compatibility, and one chunk exceeds the default size threshold; see the note on chunking below. `tsc --noEmit`, which now runs as part of this build, is clean.
- `make check-ui` passes, covering three behaviours a successful build does not demonstrate: that the server can still find the built asset paths, that a non-default language resolves and is fetchable at runtime, and that plugin translations register regardless of module evaluation order, verified through both the deferred and the immediate registration paths. All three checks fail closed: each presets its result to failure, bounds every step that can hang with its own timeout, and only reports success once it has proved it.
- Every check was observed failing before being relied on. `./script/check-built-assets.sh --self-check` rewrites the built `index.html` with four shapes the server cannot parse, including a non-stylesheet manifest link, and asserts the check fails on each, so a check that quietly stopped asserting anything is distinguishable from a passing one. The same script also fails if the Go test it depends on is ever renamed or removed, instead of silently matching nothing.
- On the Go side: `go build ./...` succeeds, so the embed still resolves; `go vet ./...` reports no issues; `TestGetStyleResolvesBuiltAssets` passes; and the license header check (`script/check-asf-header.sh`) passes.
- The built application was booted against sqlite3 and loaded in a browser. `/`, `/questions`, `/tags` and `/users` render with the client mounted, every asset and font loads, and the console is empty.

This project has no test runner and no unit tests, before or after. These are three targeted regression checks, not a test suite.

### Measurements

Same machine, same Node and package manager versions, clean tree and clean install on both sides, five runs per timing metric.

| Metric | Before | After |
|---|---|---|

- Cold and warm builds are within noise of each other, because the new build has no meaningful persistent cache to warm and now also runs a type check on every build, cold or warm alike. The old toolchain had a real warm cache, which is why its two figures differ so much. The comparison to draw is cold against cold.
- Dev server time-to-ready is wall clock on both sides, including process spawn. The new tool self-reports a much smaller number that excludes that, and using it would compare two different quantities.
- Chunk boundaries differ structurally between the two bundlers, so the bundle rows compare total shipped bytes rather than like-for-like chunks.
- HMR latency was measured once, at commit 7200ca3, on both toolchains, and carried forward from there: none of the commits since touch the hot-update path.
- All figures were captured at the branch head, d06b623, with a clean working tree on both sides.

### Not included

This PR adds no CI job for the frontend. The project runs no frontend job today, and a build on every push is a cost a maintainer should choose to take on, not one this migration should impose. One constraint carries forward for whoever wires that job later: a bare `go test ./...` reports ok while asserting nothing about the built asset paths, because `TestGetStyleResolvesBuiltAssets` skips when no frontend build is embedded. Any future CI job needs to build the frontend first.

The dev server now binds to loopback only by default; reaching it from another device on the network needs an explicit `--host` flag.

Create React App's SVG-as-component imports (`import { ReactComponent as X } from './x.svg'`) are not carried over to this configuration. Nothing in this codebase used them.


mingcheng:
This is a HUGE pull request, I don't recommend merging it without any purpose or dicusssion.

xantorres:
@mingcheng Agreed: discussion first.

The purpose: react-scripts is unmaintained and officially sunset; it's behind most of `ui/`'s audit findings and blocks dependency upgrades. Staying on it is the riskier default.

The PR just makes the discussion concrete: the server-side coupling is mapped and the risky parts are already worked out. Happy to move to a Discussion, an issue, or the dev list, and to split this into reviewable stages. Your call, no rush.

robinv8:
### Code review

Thanks for the careful write-up and the work that went into this migration — the PR description maps the server-side coupling clearly, and the fail-closed checks make the risky parts much easier to trust.

I went through this with AI-assisted review (diff + the main coupling points: Go SSR asset scraping, `header.html`, router `import.meta.glob`, env/`base`, plugin i18n init order, and the new `check-ui` guards), and ran a local smoke test of the production binary (embedded Vite build on `:9080`) and the Vite dev server (`:3001`). This is not a full line-by-line human audit of every change.

**No blocking defects found** for the default same-origin deployment from that pass. The server-side coupling fixes look correct and necessary: DOM-based `GetStyle()`, multi-stylesheet emission, and `type=module` on SSR pages match what Vite emits. Locally, the home page loads with both entry stylesheets and the module entry script, static assets and `/answer/api/v1/siteinfo` respond as expected, and the app boots without a white screen.

**Non-blocking notes for merge / rollout:**

1. **CDN + ES modules** — Moving from classic deferred scripts to `type=module` means any CDN origin must send CORS headers, or the app will fail to boot. Already called out in the PR body; please surface it in upgrade notes / CDN plugin docs, and smoke-test SSR with the CDN plugin enabled.

2. **Regression matrix still worth running** — subdirectory `base_url`, OAuth/callback URLs, `public_url` as a CDN absolute URL (rebuild UI + binary, not just restart), and non-default locales. These paths have historically been sensitive to `PUBLIC_URL` / basename changes.

3. **Process** — Agree this is a large surface. The Go and UI changes are coupled, so one PR is defensible; if maintainers prefer smaller reviews, a staged split (toolchain → SSR parsing → check scripts) is still reasonable.

4. **Optional follow-ups** — Wire `make check-ui` into CI; update comments on `GetStyle` / `check-built-assets` (they still describe tag-shape coupling, while the implementation is DOM-tolerant and fails on missing assets).

Happy to dig into a specific area if useful, or re-check after a split.

xantorres:
@robinv8 Thank you for running it locally, that's exactly the verification this needed.

1. CDN/CORS: agreed. Point me at where upgrade guidance lives and I'll add the note there, with the crossorigin rationale from the PR body.
2. Regression matrix: partially covered today (non-default locales are guarded by `check-ui` at the runtime-fetch level; `base_url` wiring is exercised through the `%BASE_URL%` manifest and the custom.css derivation), but subdirectory deploys, OAuth callbacks, and an absolute CDN `public_url` with a full rebuild are not. I'll run those three and report results here.
3. Split: maintainers' call, happy either way.
4. Stale comments on `GetStyle` / `check-built-assets`: good catch, fixed in 0f668d5. On CI: left out deliberately since a frontend build per push is a cost maintainers should opt into (the constraint for whoever wires it is in the PR body); glad to write the job if wanted.

robinv8:
@xantorres Thanks for the follow-up.

For CDN/CORS: the users who need this are those on the CDN plugins listed at https://answer.apache.org/plugins/ (`cdn-aliyun`, `cdn-s3`). A short note on the public upgrade guide (https://answer.apache.org/docs/upgrade) plus a CORS requirement in those plugin READMEs is probably enough—default same-origin deploys are unaffected. Happy to leave the exact wording to you (including the `crossorigin` rationale from the PR body).

Looking forward to the subdirectory / OAuth / absolute `public_url` results when you have them.

xantorres:
@robinv8 Ran the three deployment scenarios from the review. Summary: all three work on this branch, one behaviour change is confirmed (the CDN/CORS note already in the PR body), and three defects turned up that reproduce identically on `main`, so they are pre-existing rather than migration effects.

Method for each scenario: full rebuild of the frontend and the binary through the repo's own build path, fresh sqlite install, boot, then checks in a browser as well as over HTTP. Baseline comparisons are `main` at `3b9f137` built and booted the same way, so anything that fails on both is visible as pre-existing. Asset checks validate response bytes rather than status codes, since the SPA fallback answers 200 with HTML for unknown paths.

**1. Subdirectory deploy.** Built and booted with `ui.public_url` set to a path-suffixed origin (`http://127.0.0.1:9080/forum`) and the matching `base_url` / `api_base_url`. Server-rendered pages emit every script, stylesheet, manifest and custom.css reference under the prefix; all of them fetch with correct content types and real bytes; the module entry executes and React mounts; client-side navigation stays under the prefix (verified by setting a marker on `window` first, so a full document reload could not be mistaken for client-side routing); a direct deep link to an SPA route returns the app shell byte-identical to the built `index.html`; the browser network log shows zero 404s across 25 requests, including the on-demand route chunk and the bootstrap-icons font. Same results on `main` under the same config.

Three things fail under a subdirectory deploy on both builds equally, so they look like pre-existing issues rather than anything this PR introduces: avatar URLs are emitted as `/forum/users/https://www.gravatar.com/avatar/...`, with the base path prepended to an already-absolute URL; the 192x192 icon link is emitted with an empty `href`; and the error redirect for an unknown connector targets a root-relative `/50x` rather than `/forum/50x`. Happy to open separate issues for these if they are not already tracked, though I would rather not fold fixes into this PR.

**2. OAuth callback paths.** Compared construction rather than running a live handshake, and I want to be explicit about that boundary. All fifteen backend files on the connector and base-path surface are byte-identical between `3b9f137` and this branch. The redirect_uri is built server-side from the database `site_general.site_url` plus hardcoded route literals, in `internal/controller/connector_controller.go:131-132` and re-derived at `:148-149`; it never reads the request host or the config `ui` block, and nothing in the diff touches it. The frontend files in the oauth flow are identical too; the only nearby changes are the mechanical env-access swaps (`process.env.REACT_APP_BASE_URL` to `import.meta.env.REACT_APP_BASE_URL` for the router basename), which resolve to the same value because `envPrefix` is set and `scripts/env.js` still generates the key from the same config source.

Live probes under both root and subdirectory configs return identical results on both builds: `connector/info` returns an empty list, and unknown-connector login and redirect both 302 to `/50x`. The callback landing routes are served under the prefix on both, and loading `/forum/users/auth-landing?access_token=...` in a browser mounts the page, runs the token handler, and redirects back under the prefix.

Not exercised, and not claimed: the provider-side redirect with a real authorization code, token exchange, and `state` validation against a live-issued cache entry. The connector interface has no in-tree implementations, so those paths belong to out-of-tree plugins and cannot be reached from a stock build. Also untested are the existing-user and email-match branches of external login, and a subdirectory deploy behind a real reverse proxy rather than a direct base-path mount.

**3. Absolute CDN public_url.** Rebuilt (not just restarted) with `ui.public_url` pointing at a second local origin serving the built assets, with a CORS toggle on that origin. With `Access-Control-Allow-Origin` present: server-rendered pages emit asset URLs on the CDN origin, the module entry loads cross-origin, React mounts, the console is clean, and the icon font loads. A non-default locale also loads normally, which is worth noting because locale data comes from the application rather than the CDN, so that path is unaffected by the origin split.

With the header absent, the failure mode is worth stating precisely, because it does not look like a failure. The page still returns 200 and still renders a fully populated server-rendered question list, but React never mounts: the root element has no fiber attached, and nothing is interactive. Isolating the mechanism on the same file and origin pair, with CORS off: a classic script tag loads it, a module script tag with `crossorigin` fires an error event, and `fetch` in cors mode fails. That is the CDN/CORS note in the PR body, confirmed.

For comparison, `main` under the same config still boots without CORS headers, because its script tags are classic rather than modules. So this is a real behaviour change for CDN deployments and it is a boot-level one, not a cosmetic one. One qualifier in the other direction: the bootstrap-icons font fails on `main` too when the headers are absent, since CSS font fetches are CORS-mode requests either way. A CDN with no CORS headers was already serving broken icons before this change; what is new is that the application no longer starts.

The docs you pointed at are up as companion PRs: the upgrade-guide caution is apache/answer-website#352, and CORS sections with provider-native config examples for both CDN plugin READMEs are apache/answer-plugins#326. Both are marked best merged with or after this PR.

robinv8:
@xantorres Follow-up on docs placement.

Please put the CORS note only in the CDN plugin READMEs (`cdn-aliyun` / `cdn-s3`) — i.e. apache/answer-plugins#326.
Reason: operators fix this on the OSS/S3/CDN CORS settings, not in Answer's upgrade steps; only CDN users need this note.

Answer and plugin releases are not coupled, so:
- Plugin docs can merge first (apache/answer-plugins#326)
- When Answer ships the Vite frontend, call it out in that version's **release notes** (short note + link to the plugin README)

No need to merge the website docs PR (apache/answer-website#352) — please close it.
Do not add this to the standing upgrade guide or configfile docs.

Correcting my earlier suggestion to put this in the upgrade guide.

xantorres:
@robinv8 Done: apache/answer-plugins#326 now also covers your nit (the 200-but-not-interactive shape, 48a960a), and apache/answer-website#352 is closed. When the release that ships this is being cut, happy to draft the short release-notes line plus the plugin-README link if useful.

robinv8:
@xantorres One more process note: please retarget this PR to `dev` instead of `main` (rebase or merge the latest `dev` if needed).

We land changes on `dev` first and only promote to `main` through the normal release process. After review, this should be merged into `dev`, not directly into `main`.

xantorres:
@robinv8 Retargeted to dev. Merged the latest dev into the branch first (clean merge: dev's 15 commits touch only internal/migrations, zero overlap with this diff) and re-ran the full verification after the merge: build with the type check, all three check-ui guards, go build, go vet, and the controller test, all green. Head is bf85da6.

robinv8:
Hi @xantorres, thank you for the work on this, and for rebasing onto `dev`. Replacing CRA with Vite is something we want, and the PR description already shows you have thought through the hard parts (SSR tags, module scripts, i18n init order).

Because this touches both the UI toolchain and the Go HTML we embed, I think the next step is a tracking issue plus a few child issues, rather than continuing to review this as one change. Would you be open to opening those first? A split that should keep `dev` buildable after each step:

**Parent:** Tracking: replace Create React App with Vite (scope and constraints only)

**Child 1 — Go HTML / asset contract** (can land while CRA still builds): `GetStyle()` via `x/net/html`, `cssPath` as a list, `header.html` ranging stylesheets, plus a test. Keep classic `defer` scripts until child 2.

**Child 2 — Vite cutover** (depends on child 1; please keep this atomic): Vite config, package.json / lockfile, move `index.html`, remove CRA / rewired, env and `import.meta.env`, router `import.meta.glob`, `require()` to ESM, `header.html` `type=module`, plus the two first-paint fixes (plugin i18n init order, bootstrap-icons font path).

**Child 3 — check-ui and small follow-ups** (depends on child 2; not required to boot): the check scripts, `make check-ui`, locale-probe, and the small CSS/route nits.

I would suggest closing this PR after the tracking issue is up. This is not a rejection of the work. The commits are still on your branch, and they can come back as the child PRs. Reviewing and landing will be much easier that way. Please tweak the split if I missed something. Happy to review the series against `dev`.

xantorres:
@robinv8 Yes, happy to, and thanks for laying out the split so concretely. The branch already groups almost exactly that way. Tracking issue, the three child issues, and the first child PR (the Go asset contract, verified to land while CRA still builds) are coming shortly against dev, and I will close this one with a pointer once the tracking issue is up.

robinv8:
@xantorres Thanks again for taking this on so quickly. Looking forward to the tracking issue and the first child PR.

xantorres:
Thanks @robinv8, and agreed on the split. The series is up:

- Tracking issue: #1578
- Step 1, Go HTML and asset contract: #1579, PR #1582 (lands while Create React App still builds; classic `defer` scripts kept)
- Step 2, Vite cutover, atomic: #1580 (PR follows once step 1 lands, from `build/vite-2-cutover` on my fork)
- Step 3, check-ui and small follow-ups: #1581 (PR follows step 2, from `build/vite-3-check-ui`)

Each child branch is rebuilt on top of `dev` and its predecessor with `git cherry-pick -x`, so every commit points back at the SHA it came from here. Closing this PR as suggested; the branch stays for reference, and this thread remains the record of the review, the regression matrix, and the measurements.

PR #1575, opened 2026-08-18
fix(i18n): translate the remaining English strings in de_DE

culfin:
164 entries in `de_DE.yaml` were still the English source text. Some of them sit on pages every visitor sees:

| key | before | after |
|---|---|---|
| `question_detail.created` | Created | Erstellt am |
| `question_detail.Edited` | Edited | Bearbeitet am |
| `question_detail.Views` | Views | Aufrufe: |
| `related_question.title` | Related | Ähnliche Fragen |
| `nav_menus.ask_a_question` | Create Question | Frage stellen |

Also translated: the AI assistant panel, the badge names, the suspension and deletion notices, the review queue, and the parts of the admin interface that were still English.

## Proposed Changes

  - translate the 164 untranslated strings
  - adjust the date formats to German convention

## Date formats

Answer ships `DD. MMM` for German, which drops the year. On an archive reaching back twenty years that is not a detail — a post from 2003 and one from yesterday look alike:

```yaml
dates.long_date:            DD. MMM                 -> DD.MM.YYYY
dates.long_date_with_year:  DD. MMM YYYY            -> DD.MM.YYYY
dates.long_date_with_time:  DD. MMM YYYY [at] HH:mm -> DD.MM.YYYY [um] HH:mm
```

## Left in English on purpose

Technical terms, because that is what German speakers actually use: SMTP, SEO, API, MCP, URL, Gravatar, GitHub, HTTPS, Logo, Favicon, Status, Version, Code, Layout.

## Checks

Every placeholder was compared against `en_US.yaml` across all 1597 strings and is unchanged. The file parses, and quoting follows the existing style of each line rather than being normalised, so the diff contains translations only.

Context: I run these translations on a German optometry forum with roughly 94000 posts migrated to Answer, so all of them are in daily use.

### DIFF
--- a/i18n/de_DE.yaml
+++ b/i18n/de_DE.yaml
@@ -64,9 +64,9 @@ backend:
       user:
         other: Benutzer
       admin:
-        other: Admin
+        other: Verwaltung
       moderator:
-        other: Moderator
+        other: Moderation
     description:
       user:
         other: Standard ohne speziellen Zugriff.
@@ -235,7 +235,7 @@ backend:
       content_cannot_empty:
         other: Der Inhalt darf nicht leer sein.
       content_less_than_minimum:
-        other: Not enough content entered.
+        other: Der Text ist zu kurz.
     rank:
       fail_to_meet_the_condition:
         other: Ansehenssrang erfüllt die Bedingung nicht.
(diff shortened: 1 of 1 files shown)


mingcheng (approved):
lgtm

Issue #1578, opened 2026-08-22
Tracking: replace Create React App with Vite


xantorres:
This is the tracking issue for replacing `react-scripts` and `react-app-rewired` with Vite in `ui/`, opened at maintainer request as the follow-up to #1567. It records scope and constraints only. The work itself lands as three child pull requests, each of which keeps `dev` buildable on its own.

**Why.** `react-scripts` is unmaintained and officially sunset. It sits behind most of the `ui/` audit findings and blocks dependency upgrades, and staying on it is the riskier default. #1567 has the full write-up, the regression matrix results, and the measurements.

**The series.** Each child has its own issue and its own PR against `dev`. Order matters: 2 depends on 1, 3 depends on 2.

- [ ] Step 1, Go HTML and asset contract: #1579. `GetStyle()` reads the script and stylesheet tags out of the built `index.html` from the parsed document (`golang.org/x/net/html`) instead of a regex pinned to one bundler's tag shape; `cssPath` becomes a list; `ui/template/header.html` ranges over the stylesheets; a Go test asserts the parse against the embedded build. Lands while Create React App still builds. Classic `defer` scripts stay until step 2.
- [ ] Step 2, Vite cutover, atomic: #1580. Vite config, `package.json` and lockfile, `index.html` moved to the `ui` root, Create React App and rewired removed, `import.meta.env`, router `import.meta.glob`, `require()` to ESM, `header.html` entry script as `type="module"`, plus the two first-paint fixes (plugin i18n init order, bootstrap-icons font path).
- [ ] Step 3, check-ui and small follow-ups: #1581. The `make check-ui` guard scripts, the locale probe, and the small CSS and route fixes. Not required to boot.

**Constraints every step preserves.**

Server-side contracts. The build directory stays `ui/build`, which `ui/static.go` embeds. Emitted assets stay under `static/`, which `internal/router/ui.go` serves as a route, grouped as `static/js`, `static/css`, `static/media`. `ui/scripts/env.js` remains the single source of truth for configuration shared with the server, the `REACT_APP_` prefix included, and its `public_url` value drives the bundler's `base` so a non-root `ui.public_url` keeps every asset reference prefixed. Production sourcemaps stay on. `GetStyle()` must not depend on tag attribute order, attribute set, or quoting, so the next bundler change cannot silently empty the server-rendered pages again.

[...]

PR #1582, opened 2026-08-22
fix: parse built asset tags from the document and render every entry stylesheet

### LINKED ISSUE #1579: Vite migration (1/3): Go HTML and asset contract
Part of #1578. First step of the series; can land while Create React App still builds.

**Problem.** `internal/controller/template_controller.go` `GetStyle()` scrapes the script and stylesheet paths out of the built `ui/build/index.html` with regexes that require classic scripts with `defer` first and stylesheet links with `href` before `rel`, and it returns a single stylesheet path. Every server-rendered page reuses those paths through `ui/template/header.html`. Any other tag shape returns nothing, and the pages load with no JavaScript and no stylesheet while every build step still reports success. Nothing in the tree asserts the contract.

**Change.** Read the tags from the parsed document (`golang.org/x/net/html`), so attribute order, attribute set and quoting no longer matter. Return `cssPath` as a list, mirroring how script paths are already collected and prefixed, and range over it in `header.html`. Add `TestGetStyleResolvesBuiltAssets`, which calls the real `GetStyle()` against the real embedded build and skips when no build is embedded. Keep the classic `<script defer="defer" src=...>` tag in `header.html`; the module switch belongs to step 2.

**Done when.** `dev` with only this change builds the frontend with the current toolchain, `go build ./...` and `go vet ./...` pass, the test passes against that build, and the served pages are unchanged (one stylesheet, classic scripts).

xantorres:
Fixes #1579. Part of #1578. Step 1 of 3; lands while Create React App still builds, and step 2 (the Vite cutover) depends on it.

Three parts of the Go side had one bundler's output format encoded into them. This PR removes two of them ahead of the toolchain change; the third, the classic script tag in `header.html`, is the one line step 2 flips together with the build that emits modules, so the cutover itself stays a frontend-only diff plus that line.

**`GetStyle()` matched a literal tag shape.** It scraped `index.html` with regexes requiring classic scripts with `defer` first and stylesheet links with `href` before `rel`. Any other shape returned nothing, and server-rendered pages would load with no JavaScript and no stylesheet while every build step still reported success. The tags are now read from the parsed document (`golang.org/x/net/html`, already in the module graph and promoted from indirect to direct in `go.mod`), so attribute order, attribute set and quoting no longer matter, and the next bundler change cannot reintroduce this. The parser accepts any shape as long as the tag is present; what it guards is a missing script or stylesheet tag.

**`GetStyle()` returned a single stylesheet.** The current build emits exactly one entry stylesheet; the Vite build emits two, so pages would load partially unstyled. It is now a list, mirroring how script paths were already collected and prefixed, and `header.html` ranges over it. With today's build that is a list of one, and the rendered page is unchanged.

**`header.html` keeps its classic script tag.** `<script defer="defer" src="{{$path}}">` is untouched here on purpose; it becomes a module tag in step 2, together with the build that emits modules.

**Test.** `TestGetStyleResolvesBuiltAssets` in `internal/controller/template_controller_test.go` calls the real `GetStyle()` against the real embedded build and requires at least one script source and every declared stylesheet to come back. It skips when no frontend build is embedded, which is a constraint for anyone wiring CI later: a bare `go test ./...` reports ok while asserting nothing about the asset paths unless the frontend was built first. The self-check harness that rewrites the built `index.html` into unparseable shapes and asserts the test fails on each follows in step 3.

Agentic tooling did the mechanical work in this series; every change was reviewed by a human before being committed.

### Verification

On `dev` at `2c0ced32` plus these commits, with the current toolchain: `pnpm install --frozen-lockfile` and `pnpm build` (react-scripts) complete; `go build ./...` and `go vet ./...` pass; `TestGetStyleResolvesBuiltAssets` passes against the Create React App output (`--- PASS`, not skipped), and as a negative control the same test fails when the built `index.html` is rewritten without a script `src`, then passes again once the original is restored. The built index.html at that point carries eight classic `<script defer="defer" src="/static/js/...">` tags, one inline script and one stylesheet link, and the parser returns all eight sources and the one stylesheet.

The commits carry `(cherry picked from commit ...)` lines pointing at the branch behind #1567, where this change was first reviewed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -64,6 +64,7 @@ require (
 	go.uber.org/mock v0.6.0
 	golang.org/x/crypto v0.53.0
 	golang.org/x/image v0.20.0
+	golang.org/x/net v0.56.0
 	golang.org/x/term v0.44.0
 	golang.org/x/text v0.39.0
 	gopkg.in/gomail.v2 v2.0.0-20160411212932-81ebce5c23df
@@ -170,7 +171,6 @@ require (
 	go.uber.org/zap v1.27.0 // indirect
 	golang.org/x/arch v0.10.0 // indirect
 	golang.org/x/exp v0.0.0-20240909161429-701f63a606c0 // indirect
-	golang.org/x/net v0.56.0 // indirect
 	golang.org/x/sys v0.46.0 // indirect
 	golang.org/x/tools v0.47.0 // indirect
 	google.golang.org/protobuf v1.34.2 // indirect
--- a/internal/controller/template_controller.go
+++ b/internal/controller/template_controller.go
@@ -20,6 +20,7 @@
 package controller
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"html/template"
@@ -50,6 +51,7 @@ import (
 	"github.com/apache/answer/ui"
 	"github.com/gin-gonic/gin"
 	"github.com/segmentfault/pacman/log"
+	"golang.org/x/net/html"
 )
 
 var SiteUrl = ""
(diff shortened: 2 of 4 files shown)


LinkinStars (approved):
LGTM

PR #1587, opened 2026-08-24
fix(ui): give every page a main landmark and mark the sidebars as navigation

culfin:
No page of the interface carries a `main` landmark, and the sidebars are plain
`div`s.

### Visible on this project's own instance

`meta.answer.dev/questions`, fetched with a Googlebot user agent:

```console
$ curl -s -A "…Googlebot…" https://meta.answer.dev/questions | grep -c "<main"
0
```

Same in the source:

```console
$ grep -rn "<main\|role=\"main\"" ui/src ui/template | wc -l
0
```

### Why it matters

Screen reader users navigate by landmarks. Without a `main`, there is no way to
skip past the header and the sidebar to the content — every visit to every page
starts by tabbing through the whole navigation again. That is WCAG 2.4.1
(Bypass Blocks), and a landmark is the least intrusive way to satisfy it.

### The change

Every route now sits inside exactly one `main`:

| layout | covers |
|---|---|
| `SideNavLayout`, `SideNavLayoutWithoutFooter` | the forum, tags, users, badges |
| `Admin` | 28 admin routes |
| new `PlainLayout` (pathless) | 19 routes with no layout of their own — signing in, registering, account recovery, the error pages |

`pages/Layout` would have been the wrong place for a single landmark: it wraps
the header, so `main` there would put the navigation inside the content region
and "skip to content" would land before it.

The sidebar containers become `nav` in the same three files. Class names are
untouched, so nothing moves on screen.

### Verified

- Structurally: 88 paths before, the same 88 after, no page removed, three
  layout nodes added.
- By running it: built the front end and stepped through `/users/login`,
  `/users/register`, `/50x` and an unknown path — each renders inside exactly
  one `main` and draws its content.
- prettier, eslint, tsc.

**The router diff looks larger than it is.** Read with `git diff -w` it is 24
added lines; everything else is one indentation level.

### Deliberately not included

With two `nav` landmarks on a page — this one and the header — an `aria-label`
on each would tell them apart. That needs a new i18n key across 45 language
files, which does not belong in a change that is otherwise markup only.

Supersedes #1589 and #1591, which were the same work split across three pull
requests.

### DIFF
--- a/ui/src/pages/Admin/index.tsx
+++ b/ui/src/pages/Admin/index.tsx
@@ -36,20 +36,20 @@ const Index: FC = () => {
   });
   return (
     <div className="admin-container d-flex">
-      <div
+      <nav
         className="position-sticky px-3 border-end pt-4 d-none d-xl-block"
         id="pcSideNav">
         <AdminSideNav />
-      </div>
+      </nav>
       <div className="flex-fill w-100">
         <div className="d-flex justify-content-center px-0 px-md-4">
-          <div className="answer-container main-mx-with">
+          <main className="answer-container main-mx-with">
             <Row className="py-4">
               <Col className="page-main flex-auto">
                 <Outlet />
               </Col>
             </Row>
-          </div>
+          </main>
         </div>
         <div className="d-flex justify-content-center px-0 px-md-4">
           <div className="main-mx-with">
--- a/ui/src/pages/PlainLayout/index.tsx
+++ b/ui/src/pages/PlainLayout/index.tsx
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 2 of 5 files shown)


robinv8:
Thanks for consolidating this into one PR.

The SPA change looks right: one `main` per layout, sidebars as `nav`, and a pathless `PlainLayout` so login / error pages get a landmark without putting the header inside `main`. Plugin routes hang off `SideNavLayout`, so they pick this up too.

Two gaps worth confirming:

1. Server-rendered HTML is unchanged. `ui/template/header.html` still has `<nav id="header">` and `<div id="pcSideNav">`, and the content slot is not a `main`. The curl/Googlebot check in the description would still print 0 after this lands, because that path never goes through React. Screen-reader users who run JS are covered; crawlers and no-JS first paint are not. If the goal is only the hydrated UI, please say so in the description. If the curl case is in scope, the Go templates need the same landmarks.

2. `pages/Layout` replaces `<Outlet />` with `HttpErrorContent` when `httpStatusCode` is set, so API-driven 403/50x never reach the layout `main`.

The missing `aria-label` on the two `nav`s is fine as a follow-up (agreed that 45 locale files do not belong here).

No objection to landing the SPA markup as-is once (1) is explicit.

PR #1594, opened 2026-08-27
feat(ai): thinking mode switch, vision image input, and API host normalization fixes

lb1038678031:
### What

Three improvements to the AI feature, shipped as independent commits:

1. **fix(ai): normalize API host and harden AI model listing**
   - The models endpoint used to call `api_host + "/v1/models"` while the chat
     client auto-appends `/v1`. Configuring a base URL that already contains
     `/v1` (very common with OpenAI-compatible gateways such as SenseNova,
     SiliconFlow, OneAPI relays...) produced `/v1/v1/models` and the gateway's
     gRPC-style `{"code":5,"message":"NOT_FOUND"}` was passed through to the UI.
   - New shared helper `schema.NormalizeAPIHost()` trims slashes/whitespace,
     appends `/v1` when missing, and keeps `/v1beta/*` endpoints intact.
   - Upstream error bodies are now summarized (`error.code/message`) instead of
     being echoed verbatim; image data is never logged.
   - Stored-key lookup prefers the chosen provider before falling back to host
     matching.
   - Default provider list: Gemini points at its OpenAI-compatible endpoint;
     Anthropic removed because the backend only speaks the OpenAI protocol.

2. **feat(ai): per-provider thinking mode**
   - New `thinking_mode` switch on each provider in Admin → AI settings.
   - When enabled, a small custom `http.RoundTripper` merges
     `"enable_thinking": true` into chat completion request bodies — the
     convention used by reasoning-capable OpenAI-compatible gateways
     (DeepSeek V4, Qwen/DashScope, SenseNova, vLLM/SGLang, ...). The existing
     `reasoning_content` streaming path renders it in the bubble UI unchanged.

3. **feat(ai): vision image input behind an admin switch**
   - New `vision_enabled` switch per provider; exposed to clients via site info
     as `ai_vision_enabled`.
   - With the switch on, users can attach up to 4 PNG/JPEG/WebP images
     (≤4MB decoded each) to the first message of a turn; images travel as
     base64 data URLs or HTTPS links and are converted into MultiContent parts.
   - Conversation history stores a textual placeholder instead of raw image
     data.

### Why now / motivation

We run self-hosted Answer as a community FAQ over an OpenAI-compatible
gateway. Two concrete blockers motivated this: the model-listing 404 above,
and no way to leverage reasoning-capable models or multimodal gateways
without patching.

### Testing

- `go test ./internal/schema/ ./internal/controller/ ./internal/service/siteinfo/ ./internal/migrations/`
  — new table tests for host normalization, transport injection, and image
  validation (count/type/size).
- Manual smoke against a SenseNova OpenAI-compatible endpoint: model listing
  succeeds with both root and `/v1` base URLs; `reasoning_content` appears with
  thinking enabled; image questions answered by a vision-capable model.

### Known trade-offs

- Image attachments are not persisted in history (placeholder only) to keep
  DB schema unchanged.

---

### DIFF
--- a/internal/controller/ai_vision_test.go
+++ b/internal/controller/ai_vision_test.go
@@ -1,0 +1,81 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package controller
+
+import (
+	"encoding/base64"
+	"strings"
+	"testing"
+
+	"github.com/sashabaranov/go-openai"
+)
+
+const tinyPNG = "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg=="
+
+func dataURL() string { return "data:image/png;base64," + tinyPNG }
+
+func TestValidateAndPrepareImagesOK(t *testing.T) {
+	msg, err := ValidateAndPrepareImages("看这张图", []string{dataURL()})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(msg.MultiContent) != 2 {
+		t.Fatalf("expect 2 parts got %d", len(msg.MultiContent))
+	}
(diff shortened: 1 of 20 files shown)


internal/controller/ai_vision_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"encoding/base64"
+	"strings"
+	"testing"
+
+	"github.com/sashabaranov/go-openai"
+)
+
+const tinyPNG = "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg=="
```
LinkinStars: Important reminder: This key should not be uploaded. Please revoke it as soon as possible and deactivate it.

LinkinStars: @lb1038678031

lb1038678031: Thanks for the careful review! That line is not a credential — it's the base64
encoding of a standard **1×1 pixel PNG** (69 bytes in total), a placeholder image
commonly used in unit tests. Nothing real was leaked, so there is nothing to revoke.

That said, I agree a long base64 blob in the source looks alarming, so I pushed
`158516b` which constructs the placeholder PNG at runtime from its signature bytes —
the source no longer contains any base64 constant that could be mistaken for a secret.
The line you commented on no longer exists in the latest diff. Thanks again for
flagging it!

Issue #1595, opened 2026-08-27
Feature clarification: AI provider enhancements — thinking mode, vision image input, and API host normalization


lb1038678031:
### Background

We run a self-hosted Apache Answer as a community FAQ site, with an
OpenAI-compatible LLM gateway (SenseNova). While integrating the AI feature we
hit three concrete gaps. Following @LinkinStars's suggestion on #1594, this
issue clarifies the proposed feature set before the PR is resubmitted.

### Proposed enhancements

**1. API host normalization (bugfix)**

The admin AI settings accept an "API host", but the model-listing endpoint
concatenates `api_host + "/v1/models"` while the chat client auto-appends
`/v1`. A base URL that already contains `/v1` — the normal convention for
OpenAI-compatible gateways (SenseNova, SiliconFlow, OneAPI relays, vLLM,
SGLang...) — therefore produces `/v1/v1/models` and fails with the gateway's
`NOT_FOUND` error, which is passed through to the UI verbatim.

Proposal: a shared `NormalizeAPIHost()` helper (trim whitespace/trailing
slashes, append `/v1` when missing, keep `/v1beta/*` endpoints intact) used by
both code paths, plus concise upstream error summaries instead of echoing raw
response bodies.

**2. Per-provider thinking mode switch**

Reasoning-capable models are increasingly common (DeepSeek V4, Qwen3,
SenseNova, ...). Proposal: an admin switch per AI provider that injects the
OpenAI-compatible `enable_thinking: true` flag into chat completion requests.
The existing `reasoning_content` streaming/rendering path already displays the
thought process, so no front-end change is required for display.

**3. Per-provider vision (image input) switch**

Proposal: an admin switch enabling image attachments in AI conversations —
up to 4 PNG/JPEG/WebP images (≤4 MB decoded each) sent as base64 data URLs or
HTTPS links, converted to MultiContent parts. The switch is exposed via site
info (`ai_vision_enabled`) so the chat input only shows the attach button for
vision-capable providers. History records keep a textual placeholder instead
of persisting image data, so no DB schema change is needed.

### Scope / non-goals

- No new provider protocols (OpenAI-compatible only); Anthropic removed from
  the default provider list since the backend speaks the OpenAI protocol.
- No multi-provider routing or per-user model selection.
- No persistence of image attachments in conversation history.

### Related

- #1594 (will be closed and resubmitted referencing this issue).
- Regarding the security reminder in #1594: the flagged constant was the base64
[...]

PR #1596, opened 2026-08-27
feat(ai): thinking mode switch, vision image input, and API host normalization fixes

lb1038678031:
Resubmit of #1594, addressing @LinkinStars's feedback there. Implements #1595 (feature clarification).

### What

Three improvements to the AI feature, shipped as independent commits:

1. **fix(ai): normalize API host and harden AI model listing**
   - The models endpoint used to call `api_host + "/v1/models"` while the chat
     client auto-appends `/v1`. Configuring a base URL that already contains `/v1`
     (common with OpenAI-compatible gateways such as SenseNova, SiliconFlow,
     OneAPI relays...) produced `/v1/v1/models`, and the gateway's gRPC-style
     `NOT_FOUND` error was passed through to the UI.
    - New shared helper `schema.NormalizeAPIHost()` trims slashes/whitespace,
     appends `/v1` when missing, and keeps `/v1beta/*` endpoints intact.
    - Upstream error bodies are summarized (`error.code/message`) instead of being
     echoed verbatim; image data is never logged.
    - Stored-key lookup prefers the chosen provider before falling back to host
     matching.
    - Default provider list: Gemini points at its OpenAI-compatible endpoint;
     Anthropic removed because the backend only speaks the OpenAI protocol.

2. **feat(ai): per-provider thinking mode**
   - New `thinking_mode` switch on each provider in Admin → AI settings.
   - When enabled, a custom `http.RoundTripper` merges `"enable_thinking": true`
     into chat completion request bodies — the convention used by
     reasoning-capable OpenAI-compatible gateways (DeepSeek V4, Qwen/DashScope,
     SenseNova, vLLM/SGLang, ...). The existing `reasoning_content` streaming
     path renders the thought process in the bubble UI unchanged.

3. **feat(ai): vision image input behind an admin switch**
   - New `vision_enabled` switch per provider; exposed to clients via site info
     as `ai_vision_enabled`.
   - Users can attach up to 4 PNG/JPEG/WebP images (≤4 MB decoded each) to the
     first message of a turn; images travel as base64 data URLs or HTTPS links
     and are converted into MultiContent parts.
   - Conversation history stores a textual placeholder instead of raw image data
     (no DB schema change).

### Security note (from #1594)

The base64 constant flagged during review was the encoding of a standard 1×1
pixel PNG placeholder used in a unit test, not a credential. To avoid any
misreading, `158516b` constructs the placeholder at runtime from the canonical
PNG signature bytes; the source no longer contains any base64 blob.

### Testing

- `go test ./internal/schema/ ./internal/controller/ ./internal/service/siteinfo/ ./internal/migrations/`
  — new table tests for host normalization, transport injection, and image
  validation (count/type/size).
- Manual smoke against a SenseNova OpenAI-compatible endpoint: model listing
  succeeds with both root and `/v1` base URLs; `reasoning_content` appears with
  thinking enabled; image questions answered by a vision-capable model.

### DIFF
--- a/i18n/en_US.yaml
+++ b/i18n/en_US.yaml
@@ -869,6 +869,9 @@ ui:
     ask_placeholder: Ask a question
     thinking: Thinking…
     thoughts: Thoughts
+    image_upload: Add images
+    image_too_large: Image exceeds the 4MB limit
+    image_max_count: You can attach up to 4 images
   notifications:
     title: Notifications
     inbox: Inbox
@@ -2360,6 +2363,12 @@ ui:
       model:
         label: Model
         msg: Model is required
+      thinking_mode:
+        label: Deep thinking mode
+        tip: Sends the enable_thinking flag so reasoning-capable models show their thought process.
+      vision_enabled:
+        label: Allow image input
+        tip: Let users attach up to 4 PNG/JPEG/WebP images (max 4MB each) to AI conversations.
       add_success: AI settings updated successfully.
     conversations:
       topic: Topic
--- a/i18n/zh_CN.yaml
+++ b/i18n/zh_CN.yaml
@@ -856,6 +856,9 @@ ui:
     copy: 复制
     ask_a_follow_up: 提出后续问题
     ask_placeholder: 提问
+    image_upload: 添加图片
+    image_too_large: 图片超过 4MB 限制
+    image_max_count: 最多只能添加 4 张图片
   notifications:
     title: 通知
     inbox: 收件箱
@@ -2318,6 +2321,12 @@ ui:
       model:
         label: 模型
         msg: 模型是必需的
+      thinking_mode:
+        label: 深度思考模式
+        tip: 开启后请求将携带 enable_thinking 参数，支持推理的模型将输出思考过程。
+      vision_enabled:
+        label: 允许图片输入
+        tip: 允许用户在与 AI 对话时附带最多 4 张 PNG/JPEG/WebP 图片（每张不超过 4MB）。
       add_success: AI 设置更新成功。
     conversations:
       topic: 主题
--- a/internal/controller/ai_controller.go
+++ b/internal/controller/ai_controller.go
@@ -111,6 +111,9 @@ type ChatCompletionsRequest struct {
 type Message struct {
 	Role    string `json:"role" binding:"required"`
 	Content string `json:"content" binding:"required"`
+	// Images carries optional attachments (base64 data URLs or HTTPS URLs)
+	// for vision-capable providers. Only the first user message accepts them.
+	Images []string `json:"images,omitempty"`
 }
 
 type ChatCompletionsResponse struct {
@@ -161,6 +164,9 @@ type ConversationContext struct {
 	Messages          []*ai_conversation.ConversationMessage
 	IsNewConversation bool
 	Model             string
+	// Images holds the attachments of the current user turn only; history
+	// records keep a textual placeholder instead of persisting image data.
+	Images []string
 }
 
 func (c *ConversationContext) GetOpenAIMessages() []openai.ChatCompletionMessage {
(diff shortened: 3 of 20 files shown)


LinkinStars (changes requested):
Thanks for the substantial work here. I found several blocking issues that should be addressed before merge.

1. **Image validation/logging can be bypassed through later messages.** The handler validates and redacts only `messages[0].images`. A signed-in user can put a large base64 data URL in `messages[1].images`: it is neither validated nor sent to the model, but the full request is JSON-marshaled into the server log. Because `ShouldBind` runs before the per-image size check and this route does not apply `http.MaxBytesReader`, this also permits avoidable memory and log-volume DoS. Please cap the complete request body before binding, reject `images` on all non-first messages, and redact attachments across every message before logging.

2. **`enable_thinking` is not a provider-neutral OpenAI-compatible parameter.** The transport unconditionally injects top-level `enable_thinking: true` for every configured provider/model. This matches DashScope/Qwen compatibility mode, but it does not match DeepSeek's documented Chat Completions API. DeepSeek documents `thinking: {"type": "enabled"}` (plus `reasoning_effort`), not `enable_thinking`; its documentation also specifies how `reasoning_content` must be preserved when tools are used. The PR description specifically names DeepSeek, so direct DeepSeek configurations will have an unsupported or ignored switch. Please make this provider/model-specific (or expose a safely validated extra-body configuration) rather than injecting one gateway-specific field globally.

Official references:
- DeepSeek Thinking Mode: https://api-docs.deepseek.com/guides/thinking_mode/
- DashScope/Qwen `enable_thinking`: https://help.aliyun.com/en/model-studio/qwen-api-via-dashscope

3. **The Gemini default-host fix does not reach existing installations.** This changes init data and the already-released AI migration, but adds no new migration. Existing deployments retain `https://generativelanguage.googleapis.com` in their `ai_config.provider` row, which normalizes to the still-invalid `/v1` path rather than Gemini's OpenAI-compatible `/v1beta/openai` path. Please add a forward migration that updates only the unchanged legacy default, without overwriting administrator-customized hosts. The new default itself is correct per Gemini's documentation: https://ai.google.dev/gemini-api/docs/openai

4. **Historical `reasoning_content` is discarded on a later turn even though tools are always included.** Conversation records persist reasoning content, but rebuilding `ConversationContext` and converting it to OpenAI messages copies only `Role` and `Content`. DeepSeek explicitly requires historical reasoning content to be passed back in tool-bearing requests. Please preserve it end-to-end when restoring records and building subsequent requests.

There is also a smaller validation gap: MIME prefixes plus base64 decoding do not prove that an attachment is a PNG/JPEG/WebP; the test intentionally uses a non-decodable PNG-shaped byte sequence. Please validate actual image content (at least magic bytes, preferably decoder/config validation) before forwarding it.

LinkinStars (changes requested):
Re-review of 8de3fbdf: the previous blocking points are addressed well (request-body cap/redaction, a forward-only Gemini migration, provider-specific DeepSeek payload, reasoning-content restoration, and image signature checks). Thank you.

One compatibility issue remains in `thinkingParamForHost`: it recognizes only hosts containing `dashscope.aliyuncs.com`. Alibaba Cloud Model Studio also documents the regional OpenAI-compatible endpoints as `https://{WorkspaceId}.{region}.maas.aliyuncs.com/compatible-mode/v1` (for example, `cn-beijing`, `ap-southeast-1`, and `us-east-1`). Those endpoints use the same top-level `enable_thinking` parameter, but the current code returns nil for them, so enabling the admin toggle silently does nothing for a valid DashScope/Model Studio configuration.

Please parse the configured URL and recognize the documented `*.maas.aliyuncs.com` hosts as well (ideally match hostnames rather than substring matching). Add coverage for a regional URL. Official reference: https://help.aliyun.com/en/model-studio/visual-reasoning

PR #1600, opened 2026-08-30
feat(cli): add URL prefix fix command

Octobug:
```sh
go run cmd/answer/main.go fix --help
Fix stored URL prefixes for branding, avatars, and post content.

WARNING: Stop Answer and back up your database before running this command.

Rows already using DST_PREFIX are skipped so the command can be rerun safely.
If a text field contains both prefixes, it will be skipped and reported.

After fixing branding data, the related site info cache will be invalidated.

Usage:
  answer fix [all|branding|avatar|post] SRC_PREFIX DST_PREFIX [flags]

Examples:
  # Preview what would change
  answer fix all /uploads/ /uploads/new/ --dry-run

  # Fix up to 50 rows
  answer fix all /uploads/ /cdn/ --limit 50

  # Fix only avatars
  answer fix avatar /uploads/ /cdn/

Flags:
      --dry-run     show what would be changed without modifying the database
  -h, --help        help for fix
      --limit int   stop after fixing NUM rows total across all selected fix types (0 = unlimited)

Global Flags:
  -C, --data-path string   data path, eg: -C ./data/ (default "/data/")
```

I've tested locally with fake data generated in sqlite db, including:

- 50,000 question
- 50,000 revision
- 250,000 answer
- 250,000 file records

It took about 5 minutes to complete the fixing.

3427590a0f1f000115bfed7d97c05bfc

### DIFF
--- a/internal/cli/fix.go
+++ b/internal/cli/fix.go
@@ -1,0 +1,327 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package cli
+
+import (
+	"fmt"
+	"strings"
+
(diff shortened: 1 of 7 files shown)


internal/cli/fix.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+	if matchesSrc {
+		return matchFix
+	}
+	return matchNone
+}
+
+// classifyText classifies free text that may embed URLs at marker positions.
+func classifyText(texts []string, src, dst string, markers []string) prefixMatch {
+	hasSrc := anyContainsPrefixNotShadowed(texts, src, dst, markers)
```
LinkinStars, line 237: `classifyText` aggregates prefix matches across all supplied fields. For a post, this means a record is skipped when (for example) `original_text` still contains `SRC_PREFIX` while `parsed_text` already contains `DST_PREFIX`.

This may be uncommon in normal flows, since these fields are usually updated together, but it can still occur after an interrupted/manual migration or from pre-existing inconsistent data. In that case, skipping the whole record leaves the old URL behind even though neither individual field contains both prefixes.

To stay consistent with the command description (“If a text field contains both prefixes…”), could we classify and guard each text field independently, then replace only the fields that still contain `SRC_PREFIX`?

PR #1602, opened 2026-09-01
charts: support using an existing PVC via persistence.existingClaim

skooNI:
Add a new `persistence.existingClaim` value that, when set, causes the deployment to mount the specified PVC instead of the chart-managed one and skips creation of the PVC resource entirely.

Tested using an existing PVC that supports ReadWriteMany.


LinkinStars (approved):
LGTM

